PR #4, opened 2024-09-30
Assignment 1:Oscar + David

Oscar-Ruelas:
Collaborated with @dav1dmoore

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -70,6 +70,27 @@ public class ProgramTests
         // Assert
         Assert.AreEqual(expectedString, percentage);
     }
+    
+    [TestMethod]
+    [DataRow(10, 10, "What an amazing job! You're now a trivia master!")]
+    [DataRow(8, 10, "Great job! It seems that you really know your stuff huh.")]
+    [DataRow(5, 10, "Um :| keep practicing and you'll eventually get even better.")]
+    [DataRow(3, 10, "Oh don't give up! Study up more and try again.")]
+    public void DisplayMotivationalMessage_DisplaysCorrectMessage(int numberCorrect, int totalQuestions, string expectedMessage)
+    {
+        // Arrange
+        StringWriter stringWriter = new StringWriter();
+        Console.SetOut(stringWriter); // Redirects Console output to StringWriter
+
+        // Act
+        Program.DisplayMotivationalMessage(numberCorrect, totalQuestions);
+
+        // Assert
+        string output = stringWriter.ToString().Trim(); // Captures console output
+        Assert.AreEqual(expectedMessage, output);
+        
+    }
+
 
 
     private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -1,5 +1,7 @@
 ﻿namespace PrincessBrideTrivia;
 
+//@dav1dmoore worked with @oscar-ruelas to pair program.
+
 public class Program
 {
     public static void Main(string[] args)
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -8,6 +8,7 @@ public class Program
         Question[] questions = LoadQuestions(filePath);
 
         int numberCorrect = 0;
+        int numberIncorrect = 0;
         for (int i = 0; i < questions.Length; i++)
         {
             bool result = AskQuestion(questions[i]);
@@ -15,12 +16,29 @@ public class Program
             {
                 numberCorrect++;
             }
+            else
+            {
+                numberIncorrect++;
+            }
         }
+        
+        
+        // added a message to let the user know that the trivia has been completed along with some additional stats
+        Console.WriteLine("\nYou have completed the trivia!");
+        Console.WriteLine("Correct answers: " + numberCorrect);
+        Console.WriteLine("Incorrect answers: " + numberIncorrect);
         Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
+        
+        DisplayMotivationalMessage(numberCorrect, questions.Length);
     }
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
+        /*
+         Casted the numerCorrectAnswers to a double value so we could a correct perentage value then casted
+            the result back to an int
+         */
+        
         return (int)((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
     }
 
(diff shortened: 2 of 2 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -48,6 +66,34 @@ public static bool DisplayResult(string userGuess, Question question)
...
         return false;
     }
+    
+    /*
+        Created the DisplayMotivationalMessage method as "feature request"
+        it takes the # of correct answers and divides it from the number 
+        of total answers and based on the result it displays a motivational 
+        message
+     */
+    public static void DisplayMotivationalMessage(int numberCorrect, int totalQuestions)
```
LeoMarlow, line 76: I would have done this with a switch statement only passing in the a single variable with the math already done. 

Single example

public static void DisplayMotivationalMessage(int Score)
{
  Switch(score)
case int score <= 50 : // Maybe some kind of range here between 20-50 so no double cases can occur
Console.WriteLine(" Um :| keep practicing and you'll eventually get even better.")

}

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -48,6 +66,34 @@ public static bool DisplayResult(string userGuess, Question question)
...
+    {
+        double percentage = (double)numberCorrect / totalQuestions * 100;
+
+        if (percentage == 100)
+        {
+            Console.WriteLine("What an amazing job! You're now a trivia master!");
+        }
+        else if (percentage >= 80)
```
LeoMarlow, line 84: I'm note sure how else-if works in C# I'm guessing the percentage is descending to not double print 80 and 50 ?  I believe  the solve for this is some kind of  range function.  Super neat feature tho gj - Leo

LeoMarlow (commented):
Left comments in the code gj on your guy's feature neat idea

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
dav1dmoore: @LeoMarlow 

We really appreciated the different perspective on how to assess the score. I really think we could have used the GetPercentCorrect Method to pass a pre-calculated score, but it would have taken a bit more code. 

In regard to the suggestion of using switch, C# doesn't allow you to use logic statements in the way you have case written. It generally requires a specific value. I researched your suggestion and found that with a little more code we could have employed when clauses to make logical assessments, but implementing them would be more code than simple if, else-if statements.

Great idea and I look forward to the challenge in future reviews!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
dav1dmoore: @LeoMarlow 

I haven't ever coded in C, C++ or C#. I actually had to test it just to be sure, but generally if statements, for loops, etc. function the same across most languages. It runs vertically, testing the first piece of logic and if it fails, it continues. Once the logic statement returns true, the code executes within it's scope and stops checking further conditions.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
LeoMarlow: @dav1dmoore  Did a little more research on the topic of ranges within switch statements and found this website that was pretty helpful for switches(Added at the bottom) .  With the release of C#7.0  you can put conditions in the case block.  Sorry didn't do my research first case I was just sudo coding. 

ex code: 

case int i when i>20 &&  i<=50 :
 Console.WriteLine(" Um :| keep practicing and you'll eventually get even better.")
break;

 https://hackajob.com/talent/blog/better-c-switch-statements-for-a-range-of-values?hs_amp=true

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
LeoMarlow: Ah got it thank you for the clarification! Gj man

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
dav1dmoore: @LeoMarlow 

We really appreciate the follow up! It gives us all an opportunity to learn more about base applications. Thanks Leo!

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -70,6 +70,27 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
...
+    {
+        // Arrange
+        StringWriter stringWriter = new StringWriter();
+        Console.SetOut(stringWriter); // Redirects Console output to StringWriter
+
+        // Act
+        Program.DisplayMotivationalMessage(numberCorrect, totalQuestions);
+
+        // Assert
+        string output = stringWriter.ToString().Trim(); // Captures console output
```
BenjaminMichaelis, line 89: What is the trim for in this case? This seems to be because of a space maybe at the beginning of one of your `WriteLine`'s?

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -70,6 +70,27 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
         // Assert
         Assert.AreEqual(expectedString, percentage);
     }
+    
+    [TestMethod]
+    [DataRow(10, 10, "What an amazing job! You're now a trivia master!")]
+    [DataRow(8, 10, "Great job! It seems that you really know your stuff huh.")]
+    [DataRow(5, 10, "Um :| keep practicing and you'll eventually get even better.")]
+    [DataRow(3, 10, "Oh don't give up! Study up more and try again.")]
+    public void DisplayMotivationalMessage_DisplaysCorrectMessage(int numberCorrect, int totalQuestions, string expectedMessage)
```
BenjaminMichaelis, line 79: nit: Test name should generally follow `MethodUnderTest_ConditionUnderTest_ExpectedResult`

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -8,20 +10,38 @@ public static void Main(string[] args)
...
             }
+            else
+            {
+                numberIncorrect++;
+            }
         }
+        
+        
+        // added a message to let the user know that the trivia has been completed along with some additional stats
+        Console.WriteLine("\nYou have completed the trivia!");
```
BenjaminMichaelis, line 29: nit: `Environment.NewLine` over `\n`
```suggestion
        Console.WriteLine($"{Environment.NewLine}You have completed the trivia!");
```

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -8,20 +10,38 @@ public static void Main(string[] args)
...
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        /*
+         Casted the numerCorrectAnswers to a double value so we could a correct perentage value then casted
+            the result back to an int
+         */
+        
+        return (int)((double)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
BenjaminMichaelis, line 44:
```suggestion
        return ((float)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
I think this would be a bit simpler instead of double casting - and the int seems unnecessary nonetheless

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -48,6 +68,34 @@ public static bool DisplayResult(string userGuess, Question question)
...
+    
+    /*
+        Created the DisplayMotivationalMessage method as "feature request"
+        it takes the # of correct answers and divides it from the number 
+        of total answers and based on the result it displays a motivational 
+        message
+     */
+    public static void DisplayMotivationalMessage(int numberCorrect, int totalQuestions)
+    {
+        double percentage = (double)numberCorrect / totalQuestions * 100;
```
BenjaminMichaelis, line 80: Why not call your `GetPercentCorrect` method?

PR #8, opened 2024-10-01
Assignment1: Leo Marlow + Parker Brown

Pbrown34:
@Leomarlow was my partner for this assignment

re-pr'd due to comparing intellitect main branch instead of Assignment 1

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -71,7 +71,18 @@ public class ProgramTests
         Assert.AreEqual(expectedString, percentage);
     }
 
+    [TestMethod]
+    [DataRow(7, new int[] {0,1,2,3,4,5,6})]
+    public void QuestionRandomizer_ReturnsExpectedQuestions(int numberOfQuestions, int[] expectedResult)
+    {
+        //Arrange
 
+        //Act
+        int[] numQuestions = Program.QuestionRandomizer(numberOfQuestions);
+
+        //Assert
+        CollectionAssert.AreEquivalent(expectedResult, numQuestions);
+    }
     private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
     {
         for (int i = 0; i < numberOfQuestions; i++)
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ public class Program
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return ((int)((double)numberCorrectAnswers / numberOfQuestions * 100)).ToString() + "%";
+        return ((int)((double)numberCorrectAnswers / numberOfQuestions * 100)) + "%";
     }
 
     public static bool AskQuestion(Question question)
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -21,7 +21,7 @@ public class Program
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return (numberCorrectAnswers / (double)numberOfQuestions * 100).ToString("F2") + "%";
     }
 
     public static bool AskQuestion(Question question)
@@ -39,7 +39,7 @@ public class Program
 
     public static bool DisplayResult(string userGuess, Question question)
     {
-        if (userGuess == question.CorrectAnswerIndex)
+        if (int.TryParse(userGuess, out int userGuessIndex) && userGuessIndex - 1 == question.CorrectAnswerIndex)
         {
             Console.WriteLine("Correct");
             return true;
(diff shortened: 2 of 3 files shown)


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -71,7 +71,18 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
...
+    public void QuestionRandomizer_ReturnsExpectedQuestions(int numberOfQuestions, int[] expectedResult)
+    {
+        //Arrange
 
+        //Act
+        int[] numQuestions = Program.QuestionRandomizer(numberOfQuestions);
+
+        //Assert
+        CollectionAssert.AreEquivalent(expectedResult, numQuestions);
+    }
```
AustinH5544, line 85: If I am not mistaken, I believe this test might just check that all elements are present, I don't think it is checking for randomness. One way to test for randomness would be to use the method that Mark and Ben created in the test file called GenerateQuestionFile() to generate sample questions. Then you can randomize it using your methods. Then you can use Enumerable.SequenceEqual() to see if they are in different orders (make sure you are only comparing the Text field though, because the answers for each question object are different)

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -21,7 +24,7 @@ public static void Main(string[] args)
 
     public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
     {
-        return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+        return ((int)((double)numberCorrectAnswers / numberOfQuestions * 100)).ToString() + "%";
```
AustinH5544, line 27: It might not be necessary to convert it to a String here using toString() method. I tried it without the toString() at the end at it worked the same. The percent sign at the end causes it to return as a string.

AustinH5544 (commented):
Assignment Details
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔
Extra Credit
Come up with a new feature for the application
Issue 3: Add a new feature request implemented ✔
Issue 3: Feature request unit tested ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -91,4 +95,17 @@ public static Question[] LoadQuestions(string filePath)
...
     }
-}
+    public static int[] QuestionRandomizer(int numberOfQuestions) 
+    {
+        int[] randomArray = new int[numberOfQuestions];
+
+        for (int i = 0; i < numberOfQuestions; i++)
+        {
+            randomArray[i] = i;   
+        }
```
dav1dmoore, line 105: This was a great addition to the original program. It runs in O(n), so in means of efficiency I couldn't add much but I might suggest instead of initializing the array and then brute forcing the filling of the array consider the below:

int[] randomArray = Enumerable.Range(0, numberOfQuestions).ToArray();

This would allow you to create an array of numbers from 0 to the numberOfQuestions-1. Following this you could continue with the Shuffle in-place followed by returning randomArray. This would still run in O(n), but would reduce your lines of code to one.

Really great job!

dav1dmoore:
Assignment Details
Pull Request title includes "Assignment 1" in the title ✔
Partner has also made a commit ✔
Pull Request targets Assignment 1 branch ✔
Issue 1: Application no longer crashes ✔
Issue 2: Unit test properly passes ✔
Extra Credit
Come up with a new feature for the application
Issue 3: Add a new feature request implemented ✔
Issue 3: Feature request unit tested ✔

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
LeoMarlow: This is a great idea. Ill try to implement this and re-comment when changes are made.  Tyty

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
LeoMarlow: You are correct about the test being just to check if elements are present. I do think that testing the randomization is a great idea for a test; its just a little out of my scope of knowledge. I did a little digging into Random.shared.shuffle and found out that it is possible to feed this a seed to produce the desired array which I could check against the array. However I do believe that testing the randomness is more of a test of Random.shared.shuffle than it is the code I wrote.  

I've never used Enumerable before but I will give it a try.  Since it is an array do you just access the field with [i] ? Any ways thanks austin good idea man!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
LeoMarlow: Ah appending to a string would cause this to be become a string.   Good catch man will fix.

LeoMarlow: Fixed by @Pbrown34

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
LeoMarlow: Tried this but couldn't get it working gonna stick with what I know and keep this code how it is.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -1,4 +1,7 @@
-﻿namespace PrincessBrideTrivia;
+﻿using Microsoft.VisualBasic;
+using System.Security.Cryptography;
```
BenjaminMichaelis, line 2: nit: What are these here for? I don't see what they are being used for.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -71,7 +71,18 @@ public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuess
         Assert.AreEqual(expectedString, percentage);
     }
 
+    [TestMethod]
+    [DataRow(7, new int[] {0,1,2,3,4,5,6})]
```
BenjaminMichaelis, line 75: I'm not seeing the advantage of making this a `DataRow` when there is only one piece of data, until there is more than one, keeping it within the method is more clear/concise

PR #16, opened 2024-10-08
Assignment2: Eric Vo + Jake Lucas

tienv9:
Assignment2: Eric Vo + Jake Lucas collaboration
@JacobLucas02

### DIFF
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -16,10 +16,42 @@ public class BaseLoggerMixinsTests
         // Act 
         // added ! to suppress null warning since they should be null
         BaseLoggerMixins.Error(null!, "");
-        BaseLoggerMixins.Warning(null!, "");
-        BaseLoggerMixins.Information(null!, ""); 
-        BaseLoggerMixins.Debug(null!, "");
+        // Assert
+    }
 
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentNullException))]
+    public void Warning_WithNullLogger_ThrowsException()
+    {
+        // Arrange
+
+        // Act 
+        // added ! to suppress null warning since they should be null
+        BaseLoggerMixins.Warning(null!, "");
+        // Assert
+    }
+
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentNullException))]
+    public void Information_WithNullLogger_ThrowsException()
+    {
+        // Arrange
+
+        // Act 
+        // added ! to suppress null warning since they should be null
+        BaseLoggerMixins.Information(null!, "");
+        // Assert
+    }
+
+    [TestMethod]
+    [ExpectedException(typeof(ArgumentNullException))]
+    public void Debug_WithNullLogger_ThrowsException()
+    {
+        // Arrange
+
+        // Act 
+        // added ! to suppress null warning since they should be null
+        BaseLoggerMixins.Debug(null!, "");
         // Assert
     }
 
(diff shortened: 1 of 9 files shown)


Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -13,8 +13,12 @@ public void Error_WithNullLogger_ThrowsException()
...
         // Arrange
 
-        // Act
-        //BaseLoggerMixins.Error(null, "");
+        // Act 
+        // added ! to suppress null warning since they should be null
+        BaseLoggerMixins.Error(null!, "");
+        BaseLoggerMixins.Warning(null!, "");
+        BaseLoggerMixins.Information(null!, ""); 
+        BaseLoggerMixins.Debug(null!, "");
```
dpalmer10, line 21: With how this test is setup, lines 19-21 are never executed, as line 18 throws a NullException and the test ends

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,9 +1,69 @@
...
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        _filePath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location) + "\\log.txt";
+        _logger = new FileLogger(_filePath, "Test");
+    }
+
+    [TestMethod]
+    public void LogFactory_CreateLogger_ReturnsFileLogger()
```
dpalmer10, line 23: Consider changing "LogFactory" to "ConfigureFileLogger" because that is the method being tested

dpalmer10 (commented):
There is an existing BaseLogger class. It needs an auto property to hold a class name. This property should be set in the LogFactory using an object initializer. ❌✔
Create a FileLogger that derives from BaseLogger. It should take in a path to a file to write the log message to. When its Log method is called, it should append messages on their own line in the file. The output should include all of the following:

    The current date/time ✔
    The name of the class that created the logger ✔
    The log level ✔
    The message ✔
    The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message"

The LogFactory should be updated with a new method ConfigureFileLogger. This should take in a file path and store it in a private member. It should use this when instantiating a new FileLogger in its CreateLogger method. ✔
If the file logger has not be configured in the LogFactory, its CreateLogger method should return null. ✔
Inside of BaseLoggerMixins implement extension methods on BaseLogger for

    Error, ✔
    Warning, ✔
    Information, and ✔
    Debug. ✔ Each of these methods should take in a string for the message, as well as a parameter array of arguments for the message. Each of these extension methods is expected to be a shortcut for calling the BaseLogger.Log method, by automatically supplying the appropriate LogLevel. These methods should throw an exception if the BaseLogger parameter is null. There are a couple example unit tests to get you started.

Use the nameof() operator when identifying the class name to the logger ✔
Ensure you turn on Warnings as Errors (TreatWarningsAsErrors) ✔
Ensure that you turn on code analysis (EnableNETAnalyzers) ✔
Turn on Nullability (Nullable) ✔
    All of the above should be unit tested.

Extra Credit

    Implement an additional logger. This logger must be unit tested. Some options to consider could be one that uses System.Console or System.Diagnostics.Trace ❌
    Implement the factory pattern using static abstract methods on a Logger interface instead.

Logger.Tests/BaseLoggerMixinsTests.cs
tienv9: Thanks so much for pointing that out! I totally didn't notice.

Logger.Tests/LogFactoryTests.cs
tienv9: Thanks for pointing that out! It should now be fixed with the correct naming format.

PR #27, opened 2024-10-15
Assignment3: Eric + Austin

AustinH5544:
@theroice
@AustinH5544


LeoMarlow (commented):
Really nice job learned some stuff reading your code. 

Create two interfaces.
 One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔
Fundamentals
Be sure you enable:
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)? ✔
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.   ✔

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,44 @@
...
+        DisplayJokes = displayJokes;
+    }
+
+    public void TellJoke()
+    {
+        string joke;
+        do
+        {
+            joke = JokeService.GetJoke();
+        } while (joke.Contains("Chuck", StringComparison.OrdinalIgnoreCase) ||
```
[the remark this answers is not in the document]
theRoice: We didn't either until we wrote this line! We realized some of the jokes have lower case "chuck" and "norris" and were looking for a more elegant way to handle filtering them out!

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IJokeService
```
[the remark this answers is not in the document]
LeoMarlow: I was wrong! fixed on my end as well

CanHazFunny/CanHazFunny/IDisplayJokes.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CanHazFunny;
+
+public interface IDisplayJokes
```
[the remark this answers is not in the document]
theRoice: See above for my reasoning about why we decided not to do this!

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,167 @@
...
+        IJokeService jokeService = new JokeService();
+        IDisplayJokes displayJokes = new DisplayJokes();
+
+        // Act
+        Jester jester = new(jokeService, displayJokes);
+
+        // Assert
+        Assert.Equal(expectedTypeOfJokeService, jester.JokeService.GetType().ToString());
+        Assert.Equal(expectedTypeOfDisplayJokes, jester.DisplayJokes.GetType().ToString());
+    }
```
AndrewKorchemniy, line 104: This was technically covered by the other tests. Tests should be focusing on business logic (behavior).

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,167 @@
...
+        IJokeService jokeService = new JokeService();
+        IDisplayJokes displayJokes = new DisplayJokes();
+        Jester jester = new(jokeService, displayJokes);
+
+        // Act
+        string joke = jester.JokeService.GetJoke();
+
+        // Assert
+        Assert.NotNull(joke);
+        Assert.NotEmpty(joke);
```
AndrewKorchemniy, line 119: What is the expected value?

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,167 @@
...
+            IDisplayJokes displayJokes = new DisplayJokes();
+            Jester jester = new(jokeService, displayJokes);
+
+            //act
+            jester.TellJoke();
+            Console.Out.Flush();
+        }
+        //assert
+        string output = consoleOutput.ToString().Trim();
+        Assert.NotEmpty(output);
```
AndrewKorchemniy, line 140: What is the expected value? We should either mock IDisplayJokes to return a test value, or create a test implementation of IDisplayJokes to return a configurable test value.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,167 @@
...
+                Jester jester = new(jokeService, displayJokes);
+
+                // Act
+                jester.TellJoke();
+                Console.Out.Flush();
+            }
+            // Assert
+            string output = consoleOutput.ToString().Trim();
+            Assert.DoesNotContain("Chuck", output);
+            Assert.DoesNotContain("Norris", output);
```
AndrewKorchemniy, line 164: Not entirely sure if this will actually work since it looks to be using the real JokeService(). If so, this will fail randomly. Read above comment.

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // Arrange
+        var jokeService = new Moq.Mock<IJokeService>();
+        jokeService.Setup(x => x.GetJoke()).Returns("Test Joke haha");
+
+        // Act
+        var joke = jokeService.Object.GetJoke();
+
+        // Assert
+        Assert.Equal("Test Joke haha", joke);
```
AndrewKorchemniy, line 20: This is testing the mocked object, and not any implementation logic.

CanHazFunny/CanHazFunny.Tests/JokeServiceTests.cs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    [Fact]
+    public void FormatJoke_NullInput_ThrowsException()
+    {
+        // Assert
+        Assert.Throws<ArgumentNullException>(() => JokeService.FormatJoke(null!));
```
AndrewKorchemniy, line 27: FormatJoke() should be private if not exposed by the interface. FormatJoke doesn't expose a nullable parameter either, so this test seems redundant.

CanHazFunny/CanHazFunny/IDisplayJokes.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    public void OutputJoke(string joke);
```
AndrewKorchemniy, line 11:
```suggestion
    void OutputJoke(string joke);
```
Interface members are public by default.

CanHazFunny/CanHazFunny/IJokeService.cs
```diff
@@ -0,0 +1,12 @@
...
+{
+    public string GetJoke();
```
AndrewKorchemniy, line 11:
```suggestion
    string GetJoke();
```
Interface members are public by default.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,43 @@
...
+        get => _jokeService!;
+        set => _jokeService = value ?? throw new ArgumentNullException(nameof(value));
+    }
+
+    private IDisplayJokes? _displayJokes;
+    public IDisplayJokes DisplayJokes
+    {
+        get => _displayJokes!;
+        set => _displayJokes = value ?? throw new ArgumentNullException(nameof(value));
+    }
```
AndrewKorchemniy, line 23: The services should ideally be readonly and set once within the constructor.
The null check can be moved into the constructor as well.

CanHazFunny/CanHazFunny/JokeService.cs
```diff
@@ -1,14 +1,36 @@
...
+    public static string ExtractJoke(JsonNode? jokeNode)
+    {
+        if (jokeNode?["joke"] is JsonNode joke)
+        {
+            return joke.ToString();
+        }
+        else
+        {
+            return "No joke found!";
+        }
```
AndrewKorchemniy, line 33: This can be simplified to de-serialize into an object directly.
Avoiding JsonNode - 
https://stackoverflow.com/questions/11260631/how-to-convert-json-text-into-objects-using-c-sharp

CanHazFunny/CanHazFunny/Program.cs
```diff
@@ -1,10 +1,12 @@
...
-class Program
+public class Program
 {
-    static void Main(string[] args)
+    public static void Main(string[] args)
     {
         //Feel free to use your own setup here - this is just provided as an example
         //new Jester(new SomeReallyCoolOutputClass(), new SomeJokeServiceClass()).TellJoke();
+
+        //new Jester(new JokeService(), new DisplayJokes()).TellJoke();
```
AndrewKorchemniy, line 10: This probably shouldn't be commented.

CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,167 @@
...
+        }
+        //assert
+        string output = consoleOutput.ToString().Trim();
+        Assert.NotEmpty(output);
+    }
+
+    [Fact]
+    public void JesterTellJoke_ScreensForChuckNorris_NoChuckJokes()
+    {
+        for (int i = 0; i < 3; i++)
```
AndrewKorchemniy, line 146: If we _really_ had no option, I would bump this to 100-1000 just to make sure.

PR #28, opened 2024-10-15
Assignment3: Jake Lucas & Ben Gales

JacobLucas02:
Assignment3: Jake Lucas and Ben Gales collaboration

@webnoben

### DIFF
--- a/CanHazFunny/CanHazFunny.Tests/IJokeServiceTests.cs
+++ b/CanHazFunny/CanHazFunny.Tests/IJokeServiceTests.cs
@@ -1,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using Moq;
+using Xunit;
+
+namespace CanHazFunny.Tests;
+
+public class IJokeServiceTests
+{
+    [Fact]
+    public void GetJoke_NullJoke_ThrowException()
+    {
+        // Arrange
+        var mockJokeService = new Mock<IJokeService>();
+        mockJokeService.Setup(js => js.GetJoke()).Throws(new ArgumentNullException());
+
+        // Assert
+        Assert.Throws<ArgumentNullException>(() => mockJokeService.Object.GetJoke());
+    }
+
+    [Fact]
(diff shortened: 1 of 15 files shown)


CanHazFunny/CanHazFunny.Tests/JesterTests.cs
```diff
@@ -1,7 +1,89 @@
+using System.IO;
```
AustinH5544, line 1: Consider making a couple tests without mock.

AustinH5544 (commented):
Assignment
For this assignment do the following:

Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔❌ **// You did unit test Jester but Code coverage could be higher.**
Fundamentals
Be sure you enable:
Extra Credit
Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)?❌✔ _**I do not believe you are doing this but I also do not know everything about mock**_
The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string.❌

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+
+namespace CanHazFunny;
+
+public class Jester
+{
+    private IJokeService _jokeService;
+    private IOutputService _outputService;
+
```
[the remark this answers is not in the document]
JacobLucas02: I forgot about that, thank you for the reminder!

CanHazFunny/CanHazFunny.Tests/IJokeServiceTests.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using Moq;
+using Xunit;
+
+namespace CanHazFunny.Tests;
+
+public class IJokeServiceTests
```
AndrewKorchemniy, line 7: We shouldn't test interfaces.

PR #37, opened 2024-10-22
Assignment4: Oscar & Leo

Oscar-Ruelas:
Collab w/ @LeoMarlow


Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,8 @@
+using Xunit;
```
AustinH5544, line 1: I know it's not officially due yet but it does say to make some tests on the storage class. It also doesn't have a check or x by it so I don't know if it is required but it is in there.

AustinH5544 (commented):
Do force any derived classes to provide an implementation for Name. ✔
Test that the equality behavior on these entities behaves as expected. ❌✔ // Kevin showed us how to do this in class today at the end(10-24)
Fundamentals
Be sure you enable:
Extra Credit
Refactor your project files to put common properties and/or items into a single file. ❌✔
Convert ILogger.CreateLogger to a generic method. ❌✔
Returns the logger classes type that is to be created. ❌✔
Has a type parameter for the configuration type ❌✔

Great job on this!! Couple minor things I saw but your code looks really good. Keep up the good work ya'll!!

Logger.Tests/StorageTests.cs
LeoMarlow: Fixed and filled

Logger/Student.cs
```diff
@@ -0,0 +1,30 @@
+namespace Logger;
+
+
```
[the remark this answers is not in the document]
LeoMarlow: Email and school year should be Assigned by default by the school and when you change majors your School year should be a new year putting you on the new catalog year. These are initialized at the constructor.

Logger/FullName.cs
```diff
@@ -0,0 +1,43 @@
...
+   Using a reference type allows the object to be passed between methods or components efficiently, 
+   avoiding unnecessary memory duplication while maintaining immutability.
+*/
+
+/*
+    The FullName record is immutable because all its properties are read-only and can only be set during
+    initialization.This ensures that once a FullName object is created, its state cannot be modified.
+*/
+
+//credits: Programming C# 12: Build Cloud, Web, and Desktop Applications by Ian Griffiths
```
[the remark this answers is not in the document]
LeoMarlow: This is actually handled in Person

Logger/Person.cs
```diff
@@ -0,0 +1,18 @@
+namespace Logger;
+
+
+
```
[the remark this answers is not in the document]
LeoMarlow: This is now fixed

Oscar-Ruelas:
Assignment4 Is NOW COMPLETE AND I HAVE SATISFIED THE REQUIREMENT TO: Test that the equality behavior on these entities behaves as expected.

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace Logger.Tests;
+    public class StorageTest
+    {
+        [Fact]
+        public void Storage_AddNewStudent_Successful()
+        {
+            // Arrange  
+            Storage storage = new();
+            IEntity student = new Student("John", "Doe", "Smith",
```
KennethWhite, line 11: Nit: Generally it is not needed to store an object in a variable of the interface or base type unless working with explicit interface implementations.

Logger/Storage.cs
```diff
@@ -1,29 +1,26 @@
...
+            return Entities.Contains(item);
+        }
+        
+        public IEntity? Get(Guid expectedGuid)
         {
-            dynamic dynamicEntity = entity;
-            return dynamicEntity.Id == expectedGuid;
-        });
+            // Using dynamic to access Id
+            return Entities.FirstOrDefault(entity => ((dynamic)entity).Id == expectedGuid);
```
KennethWhite, line 24: If all entities are derived from BaseEntity / implement IEntity there should be no need to use the `dynamic` keyword.

KennethWhite (commented):
Great job! You had fantastic test coverage and your handling of null checking was very thorough. Keep up the good work!

PR #38, opened 2024-10-23
Assignment4: Tyler + Brady

twoody0:
Partnered with @BDullanty


Logger/FullName.cs
```diff
@@ -0,0 +1,58 @@
+namespace Logger;
+
+/// <summary>
+/// We chose to make the FullName Record a Reference Type (a class over a struct), because names can change (mutable)
```
theRoice, line 4: Austin and I spent a while talking about whether or not to make this class a reference type or a value type. We initially went with reference, for the same reason you list here, but ultimately decided to make it a value type, as we decided to approach names as a whole unit instead of something that is pieced together. This class really helped put perspective into how and why design decisions are made.

Logger/IEntity.cs
```diff
@@ -2,6 +2,7 @@
 public interface IEntity
 {
 
-        // Place members here.
-
-}
+    // Place members here.
+    public string Name { get; }
+    public Guid Id { get; }
```
theRoice, line 7: The readme says to make this an init-only setter. I double checked to make sure that this line doesn't do init by default and it doesn't seem like it does unless its explicitly stated.

theRoice:
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Add an Id property of type `Guid` that is an init-only setter. ✔ I put a check mark here because its there but it doesn't have an init-only setter.
  - Do force any derived classes to provide an implementation for `Name`. ✔
- Unit test the Storage class using various entities to ensure that there are no bugs.

## Fundamentals

- Be sure you enable:
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ?

## Extra Credit

- Convert `ILogger.CreateLogger` to a generic method. ✔
  - Returns the logger classes type that is to be created. ✔
  - Has a type parameter for the configuration type ✔

theRoice:
Code looks clean and organized. Your work makes me want to start using XML in my future assignments! Lots of good tests that lead to good coverage. None of the tests struck me as extraneous.

The only real thing I noticed was not making the GUID setter in IEntity init-only. See my comment above on that. 

Really good work y'all!

Logger/FullName.cs
twoody0: Yes, I agree that this assignment definitely helped perspective into how important design decisions are, me and Brady had a lot of discussions on the different ways we should or could design the program, but I'm glad we did it the way we did.

Logger/IEntity.cs
twoody0: We made it an init-only setter, one thing to keep in mind though, is that we're not creating Id objects, on the Entity class we're creating a Guid field that sets itself to Guid.NewGuid(); which is creating a Guid on initialization. Do you think that the IEntity ID property still needs to have init?

PR #40, opened 2024-10-23
Assignment4: Ben Knoll + Kevin Flannery

BenKnoll3:
Partnered with @kflannery2001

### DIFF
--- a/Logger.Tests/BookTests.cs
+++ b/Logger.Tests/BookTests.cs
@@ -1,0 +1,125 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class BookTests
+{
+    //[Fact]
+    //public void Author_SetNull_ArgumentNullException()
+    //{
+    //    // Arrange
+    //    Book book = new Book();
+    //    // Act
+        
+    //    // Assert
+    //    Assert.Throws<ArgumentNullException>(() => book.Author = new FullName(null!, null!));
+    //}
+
+    [Fact]
+    public void Author_SetValid_Success()
+    {
+        // Arrange
+        Book book = new()
+        {
+            Author = new FullName("JK", "Rowling")
+        };
+
+        // Act
+
+        // Assert
+        Assert.Equal(new FullName("JK", "Rowling"), book.Author);
+    }
+
+    [Fact]
+    public void Title_SetValid_Success()
+    {
+        // Arrange
+        Book book = new()
+        {
+            Title = "Harry Potter 1"
+        };
+        // Act
+
+        // Assert
+        Assert.Equal("Harry Potter 1", book.Title);
+    }
+
+    [Fact]
+    public void Title_SetNull_ArgumentNullException()
+    {
+        // Arrange
+        Book book = new();
+        
+        // Act
+
+        // Assert
+        Assert.Throws<ArgumentNullException>(() => book.Title = null!);
+
+    }
+
+    [Fact]
+    public void Title_SetEmpty_ArgumentNullException()
+    {
+        // Arrange
+        Book book = new();
+
+        // Act
+
+        // Assert
+        Assert.Throws<ArgumentException>(() => book.Title = "");
+    }
+
+    [Fact]
+    public void Name_SetValid_Success()
+    {
+        // Arrange
+        Book book = new()
+        {
+            //book.Title = "Harry Potter 1";
+            //book.Author = new FullName("JK", "Rowling");
+            Name = "Harry Potter 1 by: JK Rowling"
+        };
+        // Act
+
+        // Assert
+        Assert.Equal("Harry Potter 1 by: JK Rowling", book.Name);
+    }
+
+    [Fact]
+    public void Name_SetNull_ArgumentNullException()
+    {
+        // Arrange
+        Book book = new();
+
+        // Act
+
+        // Assert
+        Assert.Throws<ArgumentNullException>(() => book.Name = null!);
+    }
+
+    [Fact]
+    public void Name_SetEmpty_ArgumentException()
+    {
+        // Arrange
+        Book book = new();
+
+        // Act
+
+        // Assert
+        Assert.Throws<ArgumentException>(() => book.Name = "");
+    }
+
+    [Fact]
+    public void Name_SetInvalidAuthor_ArgumentException()
+    {
+        // Arrange
+        Book book = new();
+        
+        // Act
+
+        // Assert
+        Assert.Throws<ArgumentException>(() => book.Name = "Harry Potter 1 by: JK ");
+    }
+
+
+}
--- a/Logger.Tests/BookTests.cs
+++ b/Logger.Tests/BookTests.cs
@@ -19,7 +19,7 @@ public class BookTests
     public void Author_SetValid_Success()
     {
         // Arrange
-        Book book = new()
+        Book book = new("Harry Potter", "JK", null, "Rowling")
         {
             Author = new FullName("JK", "Rowling")
         };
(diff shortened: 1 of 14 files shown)


Logger/Student.cs
```diff
@@ -0,0 +1,76 @@
...
+            return (int) _studentId!;
+        }
+        set
+        {
+            ArgumentNullException.ThrowIfNull(value, nameof(value));
+            _studentId = value;
+        }
+    }
+
+    // TODO: Refactor Common Members between Student and Employee
```
ccornell88, line 61: Looks like you already have a TODO to complete this.

Logger/Student.cs
```diff
@@ -0,0 +1,76 @@
...
+    private FullName? _studentName;
```
ccornell88, line 62: Already have a comment to complete.

Logger/Student.cs
ccornell88, line 61: Already have a comment to complete.

Logger.Tests/BookTests.cs
```diff
@@ -0,0 +1,125 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class BookTests
+{
+    //[Fact]
```
ccornell88, line 7: Idk if you need this commented out test or if it can be removed.

Logger/Book.cs
```diff
@@ -0,0 +1,70 @@
...
+            else if (authorNames.Length == 3) 
+            {
+                Author = new FullName(authorNames[0], authorNames[1], authorNames[2]);
+            }
+            else
+            {
+                throw new ArgumentException("Invalid Author Name", names[1]);
+            }
+        }
+    }
```
tienv9, line 39: Do force any derived classes to provide an implementation for Name. ❌

Logger/Employee.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Logger;
+
+public record class Employee : BaseEntity
+{
+    // We used explicit implementation for the Name property becuase we need to be able to take Name and parse
+    // it into the EmployeeId, EmployeeFullName, and EmployeeJobTitle properties. We also need to be able to
+    // calculate the Name property from the EmployeeId, EmployeeFullName, and EmployeeJobTitle properties.
+    override public string Name
+    {
```
tienv9, line 11: Do force any derived classes to provide an implementation for Name. ❌

tienv9 (commented):
Tasks
Do force any derived classes to provide an implementation for Name. ❌
Fundamentals
Be sure you enable:
Extra Credit
Convert ILogger.CreateLogger to a generic method. ❌
Returns the logger classes type that is to be created. ❌
Has a type parameter for the configuration type ❌

Logger/Book.cs
BenKnoll3: I am a bit confused by this comment. We believe we have implemented Name within all of our derived classes.

tienv9: Sorry, I read it wrong from the requirement above it and though it said "Do not force implementation"

Logger.Tests/BookTests.cs
KennethWhite: Commented code should be removed before opening a PR

Logger.Tests/FullNameTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    //[Fact]
+    //public void Create_FullNameWithNull_ArgumentNullException()
+    //{
+    //    // Arrange
+    //    FullName testName;
+    //    // Act
+
+    //    // Assert
+    //    Assert.Throws<ArgumentNullException>(() => testName = new FullName(null!, null!));
+    //}
```
KennethWhite, line 53: Nit: remove

Logger/Book.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+
+    public Book(string title, string authorFirstName, string? authorMiddleName, string authorLastName)
+    {
+        Title = title;
+        Author = new FullName(authorFirstName, authorMiddleName, authorLastName);
+    }
+
+    // Is it for internal or external use? implicit / explicit
+    // We used explicit implementation for the Name property becuase we need to be able to take Name and parse
```
KennethWhite, line 13: This is an implementation of an abstract interface member, but it is _not_ an explicit interface implementation. This is worth reviewing:
```
IEntity.ExplicitImplementation
vs
ImplicitImplementation
```

Logger/Book.cs
```diff
@@ -0,0 +1,77 @@
...
+            return $"{Title} by: {Author}";
+        }
+        set
+        {
+            ArgumentException.ThrowIfNullOrWhiteSpace(value, nameof(value));
+
+            string[] names;
+            names = value.Split(":");
+
+            Title = names[0].Substring(0,names[0].Length - 2).Trim();
```
KennethWhite, line 29: Having to do a lot of manual string manipulation like this usually indicates a design flaw. Consider approaches that allow you to easily print the details you want, while also allowing you to easily set the data. 

In particular you also have what we call Temporal Coupling here. You cannot set the `Name` of this object without also setting its `Title` and `Author` at the same time. If the desire is to have Name be "Title by Author.First Author.Last" then consider a computed property to avoid this coupling.

Logger/Book.cs
```diff
@@ -0,0 +1,77 @@
...
+                Author = new FullName(authorNames[0], authorNames[1], authorNames[2]);
+            }
+            else
+            {
+                throw new ArgumentException("Invalid Author Name", names[1]);
+            }
+        }
+    }
+
+    private FullName? _author;
```
KennethWhite, line 48: It doesn't make sense to declare these fields as nullable since we don't actually want to allow null values to be stored. We see this manifesting in your property getters when you needed to use the null ! operator to assert you were not returning null. 

When working in a project with nullability, generally when you use the ! operator it indicates you are not properly checking for null or that you have a nullability mismatch like we have here.

Logger/Book.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public FullName Author 
+    { 
+        get
+        {
+            return (FullName) _author!;
```
KennethWhite, line 54: This cast should not be necessary.
```suggestion
            return _author!;
```

Logger/Employee.cs
```diff
@@ -0,0 +1,71 @@
...
+    // We used explicit implementation for the Name property becuase we need to be able to take Name and parse
+    // it into the EmployeeId, PersonName, and EmployeeJobTitle properties. We also need to be able to
+    // calculate the Name property from the EmployeeId, PersonName, and EmployeeJobTitle properties.
+    override public string Name
+    {
+        get
+        {
+            return $"{EmployeeId}: {PersonName}: {EmployeeJobTitle}";
+        }
+        set
```
KennethWhite, line 23: As mentioned on the book class, setters like this are problematic. There is a lot of logic here that is not covered by unit tests as well.

Logger/Student.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    private int? _studentId;
+    public int StudentId
+    {
+        get
+        {
+            return (int) _studentId!;
+        }
```
KennethWhite, line 59:
```suggestion
    private int _studentId;
    public int StudentId
    {
        get
        {
            return _studentId;
        }
```

KennethWhite (commented):
Nice work overall. There a couple things you should review, Nullability and Implicit/Explicit interface implementations. There is some minor cleanup (commented code) that I would expect to see completed before a PR like this merges.

As I commented in the code, a few places where your setters had a lot of string logic going on. It seems like you were creating a setter for something that should just be a computed property. Then if you wish to change Title/Author rather than doing so via setting a magic string, you do it by updating those properties themselves.

Keep up the good work!

Logger/Book.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+        set
+        {
+            ArgumentNullException.ThrowIfNull(value, nameof(value));
```
KennethWhite, line 58: Also, we should avoid the use of `nameof(value)` and instead pass a more meaningful message.

PR #41, opened 2024-10-23
Assigment 3 Kasey Miller

1Kmiller1:
fixed 8.0net and got joke

### DIFF
--- a/CanHazFunny/.editorconfig
+++ b/CanHazFunny/.editorconfig
@@ -1,0 +1,8 @@
+# Remove the line below if you want to inherit .editorconfig settings from higher directories
+root = true
+
+# C# files
+[*.cs]
+
+# CA1707: Identifiers should not contain underscores
+dotnet_diagnostic.CA1707.severity = none
--- a/CanHazFunny/CanHazFunny.Tests/JokeServiceTest.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JokeServiceTest.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace CanHazFunny.Tests
+{
+    public class JokeServiceTest
+    {
+        [Fact]
+        public void ReturnNull() { }
+    }
+}
--- a/CanHazFunny/CanHazFunny.Tests/JokeServiceTest.cs
+++ b/CanHazFunny/CanHazFunny.Tests/JokeServiceTest.cs
@@ -6,25 +6,24 @@ using System.Text;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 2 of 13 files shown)


1Kmiller1:
Assignment
For this assignment do the following:

Create two interfaces. One to represent the JokeService and one interface to represent outputting the joke to the screen (Console.WriteLine). ✔
Unit test the Jester class. Code coverage should be above 90% for this class. Moq will make this significantly easier. ✔

Fundamentals

Be sure you enable:

Extra Credit

Unit test your implementation that writes the joke out to the screen. How hard could it be to unit test a single line method ;)✔

The Geek jokes API that is being used can also return jokes in a JSON format. Update the JokeService to retrieve jokes using JSON. GetJoke should still return a string. ❌

CanHazFunny/.editorconfig
AndrewKorchemniy, line 1: We shouldn't be disabling warnings unless we're certain it provides no value.

CanHazFunny/CanHazFunny.Tests/JokeServiceTest.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        string expectedJoke = "*Refrence to Other Funny Joke*";
+
+        mockjokeService.Setup(service => service.GetJoke()).Returns(expectedJoke);
+        //Act
+        string retreviedJoke = mockjokeService.Object.GetJoke();
+
+        //Assert
+        Assert.Equal(expectedJoke, retreviedJoke);
+    }
```
AndrewKorchemniy, line 27: This is testing the mocked objected and not the implementation.

CanHazFunny/CanHazFunny/Jester.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    private IJokeService _jokeService;
+    private IOutputService _outputService;
+
+    public Jester(IJokeService? jokeService, IOutputService? outputService) 
+    { 
+    
+        _jokeService = jokeService ?? throw new ArgumentNullException (nameof(jokeService));
+        _outputService = outputService ?? throw new ArgumentNullException (nameof (outputService));
+
```
AndrewKorchemniy, line 19: We shouldn't allow users to inject nullable dependencies and then immediately throw if they do.
https://stevetalkscode.co.uk/null-injection

CanHazFunny/CanHazFunny/Program.cs
```diff
@@ -1,10 +1,10 @@
 ﻿namespace CanHazFunny;
 
-class Program
+public static class Program
```
AndrewKorchemniy, line 3: Only main() itself needs to be static.

AndrewKorchemniy (commented):
late assignment

Joshua-Lester3:
Hey Kasey, just a heads up: you'll need to find a partner as that's a requirement for the assignments.

PR #42, opened 2024-10-23
Assignment4: Chris Cornell

ccornell88:


kflannery2001:
I will review this

Logger/Book.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Logger;
+
+public record Book(string Title) : BaseEntity
```
kflannery2001, line 3: DO use record class (C# 10.0) for clarity, rather than the abbreviated record-only syntax.

Logger/Employee.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Logger;
+
+public record Employee(FullName FullName) : BaseEntity
```
kflannery2001, line 3: record class

Logger/Student.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Logger;
+
+public record Student(FullName FullName) : BaseEntity
```
kflannery2001, line 3: record class

Logger/Book.cs
ccornell88: Ill get that changed. I am not quite done yet just wanted to get a pull request up.

Logger/Book.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Logger;
+
+public record class Book : BaseEntity
+
+// The Name property is implemented implicitly because the IEntity interface defines it,
+// and we override it by returning the books Title.
+{
+    public string Title { get; }
+
+    public Book(string title)
```
dpalmer10, line 10: Everything looks great, you could maybe add an author to a book so that it has a little more information available. Good work.

dpalmer10 (commented):
        Do force any derived classes to provide an implementation for Name. ✔

Fundamentals

    Be sure you enable:

Extra Credit

    Convert ILogger.CreateLogger to a generic method. ❌✔
        Returns the logger classes type that is to be created. ❌✔
        Has a type parameter for the configuration type ❌✔

Logger/Employee.cs
ccornell88: Done!

Logger/Student.cs
ccornell88: Done!

Logger/Book.cs
ccornell88: Thanks I meant to add something. ill get it done!

Logger/Book.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Logger;
+
+public record class Book : BaseEntity
+{
+    // The Name property is implemented implicitly because the IEntity interface defines it,
```
KennethWhite, line 5: Nit: Generally put comments above the property they are referring to

Logger.Tests/FullNameTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        // Act
+        var result = fullName.ToString();
+
+        // Assert
+        Assert.Equal("Chris Cornell", result);
+
+    }
+
+    [Fact]
+    public void FullName_WithBlankMiddleName_ShouldSetMiddleNameToNull()
```
KennethWhite, line 48: Nit: You could have added an additional check where the middle name was passed in as null, though this test is enough to cover that code block on its own.

KennethWhite (commented):
Overall great work! Nothing jumped out at me beyond a couple nit picky things. Keep up the good work!

Joshua-Lester3:
Hey Chris, just a heads up that you need to find a partner for future assignments. We deduct points for this.

PR #43, opened 2024-10-24
Assignment 4: Ben Gales + Jake Lucas

webnoben:


Logger/Entity.cs
```diff
@@ -0,0 +1,10 @@
+namespace Logger;
+
+public abstract record class Entity : IEntity
+{
+    // Explicit implementation
+    Guid IEntity.Id { get => throw new NotImplementedException(); init => throw new NotImplementedException(); }
```
BDullanty, line 6: Did not implement a GUID here

Logger/Entity.cs
```diff
@@ -0,0 +1,10 @@
...
+
+    // Implicit implementation
+    public abstract string Name { get; }
```
BDullanty, line 9: - Do force any derived classes to provide an implementation for `Name`. ❌
  
  By declaring name as abstract you have forced derived classes to implement name

Logger/Employee.cs
```diff
@@ -0,0 +1,15 @@
...
+    public string EmployeeId { get; init; }
+
+    public Employee(FullName fullName, string employeeId) : base(fullName)
+    {
+        EmployeeId = employeeId;
+    }
+
+    // Explicitly implementing GetId to return the EmployeeId
+    // The employee can have a unique id
+    protected override string GetId() => EmployeeId;
```
BDullanty, line 14: This is actually an implicit implementation

Logger/Student.cs
```diff
@@ -0,0 +1,15 @@
...
+    public string StudentId { get; init; }
+
+    public Student(FullName fullName, string studentId) : base(fullName)
+    {
+        StudentId = studentId;
+    }
+
+    // Explicitly implementing GetId to return the StudentId
+    // This provides a way to get the unique id for a Student
+    protected override string GetId() => StudentId;
```
BDullanty, line 14: You actually have implicit implementation here

Logger.Tests/EntityTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        var fullName2 = new FullName("Jane", "Alice", "Smith");
+        var employee1 = new Employee(fullName1, "E001");
+        var employee2 = new Employee(fullName2, "E001");
+        var employee3 = new Employee(fullName1, "E002");
+
+        // Act & Assert
+        Assert.Equal(employee1, employee2);
+        Assert.NotEqual(employee1, employee3);
+    }
+}
```
BDullanty, line 49: Need much more tests overall. Consider using test driven development
- **All of the above should be unit tested** ❌✔

BDullanty: - Unit test the Storage class using various entities to ensure that there are no bugs.❌
Missing storage class and also the tests for it.

Logger/Logger.csproj
```diff
@@ -1,12 +1,12 @@
...
-    <LangVersion>11.0</LangVersion>
-    <ImplicitUsings>enable</ImplicitUsings>
-    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+	  <TargetFramework>net8.0</TargetFramework>
+	  <IsPackable>false</IsPackable>
+	  <Nullable>enable</Nullable>
+	  <LangVersion>12.0</LangVersion>
+	  <ImplicitUsings>enable</ImplicitUsings>
+	  <EnableNETAnalyzers>true</EnableNETAnalyzers>
   </PropertyGroup>
```
BDullanty, line 11: - Ensure that you turn on `EnforceCodeStyleInBuild` ❌

BDullanty (commented):
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ✔
  - Add an Id property of type `Guid` that is an init-only setter. ✔
  - Do force any derived classes to provide an implementation for `Name`. ❌✔
- Unit test the Storage class using various entities to ensure that there are no bugs.❌

## Fundamentals

- Be sure you enable:
  - Ensure that you turn on `EnforceCodeStyleInBuild` ❌
- **All of the above should be unit tested** ❌
- **Choose simplicity over complexity** ✔

## Extra Credit

- Convert `ILogger.CreateLogger` to a generic method. ❌✔
  - Returns the logger classes type that is to be created. ❌✔
  - Has a type parameter for the configuration type ❌✔

Logger/Entity.cs
JacobLucas02: So then this code is correct?

BDullanty: Yes, my apologies, I had read it as do not.

Logger/Book.cs
```diff
@@ -0,0 +1,16 @@
+namespace Logger;
+
+public record class Book : Entity
+{
+    public string Title { get; init; }
+    public string Author { get; init; }
+
+    // Name is implemented implicitly because the book requires a specific format of it
```
KennethWhite, line 8: An implicit implementation would mean that the Book class inherits the Name property directly from the base class Entity without adding its own override.

In this case you have an explicit implementation because the Book class is defining its own specific logic for the Name property, overriding whatever might be in the base class Entity.

Logger/Employee.cs
```diff
@@ -0,0 +1,15 @@
...
+public record Employee : Person
+{
+    public string EmployeeId { get; init; }
+
+    public Employee(FullName fullName, string employeeId) : base(fullName)
+    {
+        EmployeeId = employeeId;
+    }
+
+    // Implicitly implementing GetId to return the EmployeeId
```
KennethWhite, line 12: Same comment regarding implicit/explicit.

Logger/FullName.cs
```diff
@@ -0,0 +1,36 @@
+namespace Logger;
+
+// Using 'struct' for FullName enables the use of value types
+// When comparing this class with others, its not compared by memory, but instead the contents it contains
+// Using 'record' makes this value type immutable. Meaning properties in the class are readonly by default
+public record struct FullName
+{ 
+    private string? first;
```
KennethWhite, line 8: Per the [C# guidelines](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names): Private instance fields start with an underscore (_) and the remaining text is camelCased.

Logger/FullName.cs
```diff
@@ -0,0 +1,36 @@
...
+    private string? middle;
+    private string? last;
+
+    public FullName(string? first, string? middle, string? last)
```
KennethWhite, line 12: By making all 3 parameters optional, you are essentially saying a FullName where all 3 names are null is a valid use of this Record. This is probably not desirable.

Logger/IEntity.cs
```diff
@@ -1,7 +1,9 @@
 ﻿namespace Logger;
 public interface IEntity
 {
+    public Guid Id { get; init; }
+
+    string Name { get; }
```
KennethWhite, line 6: There should be an access modifier on this property declaration.

Logger/Person.cs
```diff
@@ -0,0 +1,21 @@
+namespace Logger;
+
+public abstract record class Person : Entity
+{
+    public FullName FullName { get; init; }
+
+    // Implicitly implementing Name from Entity as a formatted string based on FullName.
```
KennethWhite, line 7: Explicit vs Implicit

Logger/Person.cs
```diff
@@ -0,0 +1,21 @@
...
+    // This implementation allows derived classes to inherit it
+    public override string Name => FullName.ToString();
```
KennethWhite, line 9: You did not provide a ToString implementation on FullName, so we get the default object ToString which is something like this:
`FullName { First = Inigo, Middle = , Last = Montoya }`

Logger/Student.cs
KennethWhite: An implicit implementation would mean that the Book class inherits the Name property directly from the base class Entity without adding its own override.

KennethWhite (commented):
There are a few things in here that could be improved upon, but generally good work here. Unit tests in particular could have been more thorough. There were equality tests for individual properties but none for the entities themselves. While there are some tests on the entities there are none that test the storage class interacts with them properly.

PR #44, opened 2024-10-27
Assignment4: David & Kasey

dav1dmoore:


JacobLucas02:
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ❌✔
  - Add an Id property of type `Guid` that is an init-only setter. ❌✔
  - Do force any derived classes to provide an implementation for `Name`. ✔
- Unit test the Storage class using various entities to ensure that there are no bugs.

## Fundamentals

- Be sure you enable:
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔

## Extra Credit

- Convert `ILogger.CreateLogger` to a generic method. ✔
  - Returns the logger classes type that is to be created. ✔
  - Has a type parameter for the configuration type ✔

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Xml.Linq;
+using Xunit;
+
+namespace Logger.Tests;
+    public class StorageTests
+    {
+    [Fact]
+    public void StudentEmployeeBookStorage_AddEntitiesToStorage_ReturnsTrue()
```
twoody0, line 8: I believe the naming convention for tests are:
ClassMethod_Input_ExpectedOutput
So I would suggest maybe something like:
Add_RecordValues_AddsToList

I like how you're testing all of the records in one test, but I wonder if it's bad coding practice because if one record causes the test to fail, then you won't immediately know which record is causing it to fail without debugging the test.

Logger.Tests/EmployeeTests.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Xunit;   
+
+namespace Logger.Tests;
+    public class EmployeeTests
+    {
```
twoody0, line 5: Nit: The EmployeeTests class as well as the { can move over to the left one tab, I recently found an extension called CodeMaid that helps with white space, and other nice QOL stuff

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+{
+2 => new FullName(nameParts[0], null, nameParts[1]),               // "First Last"
+3 => new FullName(nameParts[0], nameParts[1], nameParts[2]),       // "First Middle Last"
+_ => throw new ArgumentException("Author name must be in 'First Last' or 'First Middle Last' format.", nameof(authorName))
+};
+}
+
+
+
+}
```
twoody0, line 52: The whole white space is a little off for this file, I would highly suggest looking into that extension I mentioned earlier, at the minimum it would help with the readability

Logger/Person.cs
```diff
@@ -0,0 +1,43 @@
...
+            _ => throw new ArgumentException("Name must be in 'First Last' or 'First Middle Last' format.", nameof(fullName))
+        };
+    }
+
+    // Implement Name property from BaseEntity as a calculated property
+    public override string Name
+    {
+        get => $"{FullName.ToString()}";
+        set => throw new InvalidOperationException("The Name property is calculated and cannot be set directly.");
+    }
```
twoody0, line 42: In the BaseEntity class would it be easier if you had a ParseName method, and just have the Name property reference the ParseName method?
Then you could just override the ParseName method and simply return FullName.ToString();

Logger/Person.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+
+namespace Logger;
+    public abstract record class Person : BaseEntity
+{
+
+
+    // Common properties. ID and Name may change.
+    public FullName FullName { get; set; }
+    public int ID { get; set; }
```
twoody0, line 10: Me and my partner made the ID property be a string, because with our logic we thought that a lot of times, there will be characters in student and employee ID's, but that's more of a design question

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
...
+var nameParts = authorName.Split(' ', StringSplitOptions.RemoveEmptyEntries);
+
+return nameParts.Length switch
+{
+2 => new FullName(nameParts[0], null, nameParts[1]),               // "First Last"
+3 => new FullName(nameParts[0], nameParts[1], nameParts[2]),       // "First Middle Last"
+_ => throw new ArgumentException("Author name must be in 'First Last' or 'First Middle Last' format.", nameof(authorName))
+};
+}
+
```
twoody0, line 49: This shouldn't be necessary, because in your FullName record you already have logic for if it should return with the first, middle, last. Just call the ToString method

twoody0 (commented):
## Tasks

  - Provide a comment on the full name record on why you selected to define a value or a reference type and ❌
  - Provide a comment on the full name record on why or why not the type is immutable. ❌
  - Add an Id property of type `Guid` that is an init-only setter. ✔
  - Do not implement the `Name` property in this abstract class. ❌
  - Do force any derived classes to provide an implementation for `Name`. ✔
  - Provide a comment on each interface **member** in each entity explaining why you implemented it implicitly or explicitly. ❌
- Unit test the Storage class using various entities to ensure that there are no bugs.

## Fundamentals

- Be sure you enable:
  - Ensure that you turn on `EnforceCodeStyleInBuild` ✔
- **All of the above should be unit tested** ✔
- **Choose simplicity over complexity** ✔

## Extra Credit

- Convert `ILogger.CreateLogger` to a generic method. ✔
  - Returns the logger classes type that is to be created. ✔
  - Has a type parameter for the configuration type ✔
  
  Overall I think you did a pretty good job, I would suggest adding comments to follow the grading guidelines so you don't lose any simple points.

Logger/Person.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    // Constructor to initialize common properties
+    protected Person(string name, int id)
+    {
+        FullName = ParseName(name);
+        ID = id;
+    }
+
+    // Method to parse the full name from a single string input
+    protected static FullName ParseName(string fullName)
```
KennethWhite, line 20: This is the second implementation of this in this PR. When we encounter a situation like this we want to avoid duplication by refactoring where this lives. Two options are 
- Put this as a static method on the FullName record 
- Write a string extension method for and then use it in other files as needed

KennethWhite (commented):
Overall good work! There are some formatting issues but the code itself looks great.

Logger.Tests/BookTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        // Act & Assess
+        Assert.NotNull(newBook);
+        Assert.Equal(expected: 33 - 0987654, actual: newBook.ISBN);
+        Assert.Equal("J.K.", newBook.Author.FirstName);
+        Assert.Null(newBook.Author.MiddleName);
+        Assert.Equal("Rowling", newBook.Author.LastName);
+        Assert.Equal(2000, newBook.YearPublished);
+        Assert.Equal("Harry Potter", newBook.Title);
+
```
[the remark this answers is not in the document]
dav1dmoore: Great catch! Updated and duplicate removed.

Logger.Tests/StorageTests.cs
dav1dmoore: Updated tests for singularity. Thanks!

Logger.Tests/EmployeeTests.cs
dav1dmoore: Updated Format. Thanks!

Logger.Tests/FullNameTests.cs
```diff
@@ -0,0 +1,48 @@
...
+        FullName fullName = new("John", "Michael", "Doe");
+
+        // Act
+        var result = fullName.ToString();
+
+        // Assess
+        Assert.Equal("John Michael Doe", result);
+    }
+
+    //[Fact]
```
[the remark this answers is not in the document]
dav1dmoore: Comments removed. Thanks!

Logger.Tests/TestLogger.cs
```diff
@@ -9,9 +9,9 @@ public TestLogger(string logSource) : base(logSource) { }
     public static ILogger CreateLogger(in TestLoggerConfiguration configuration) =>
         new TestLogger(configuration.LogSource);
 
-    static ILogger ILogger.CreateLogger(in ILoggerConfiguration configuration) =>
+    static T ILogger.CreateLogger<T, U>(in U configuration) =>
         configuration is TestLoggerConfiguration testLoggerConfiguration
-            ? CreateLogger(testLoggerConfiguration)
+            ? (T)(object)CreateLogger(testLoggerConfiguration)
```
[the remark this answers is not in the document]
dav1dmoore: Removed Object cast. Thanks!

Logger/Book.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System.Reflection;
+
+namespace Logger;
+public record class Book : BaseEntity
+{
+
+    //These properties are implemented implicitly, as they belong to Book and do not require special handling.
```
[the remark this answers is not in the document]
dav1dmoore: Removed Comment.

Logger/FullName.cs
```diff
@@ -0,0 +1,34 @@
...
+                : middleName;
+
+        // LastName cannot be null or whitespace. Every person has a last name.
+        LastName = string.IsNullOrWhiteSpace(lastName)
+                ? throw new ArgumentNullException(nameof(lastName), $"'{nameof(lastName)}' cannot be null or whitespace.")
+                : lastName;
+    }
+
+
+    // Properties are already read-only and immutable due to being a record.
```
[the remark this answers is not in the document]
dav1dmoore: Removed comment. Thanks.

PR #45, opened 2024-10-29
Assignment4: Eric Vo + Daniel

tienv9:
Assignment4: Eric Vo + Daniel collaboration
@dpalmer10

### DIFF
--- a/Logger.Tests/BookTests.cs
+++ b/Logger.Tests/BookTests.cs
@@ -1,65 +1,59 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
-using Xunit;
+﻿using Xunit;
 
-namespace Logger.Tests
+namespace Logger.Tests;
+
+public class BookTests
 {
-    public class BookTests
+    [Fact]
+    public void Constructor_InitializedBook_Success()
     {
-        [Fact]
-        public void Constructor_InitializedBook_Success()
-        {
-            FullName fullName = new("William", "Goldman");
-            Book book = new(fullName, "The Princess Bride", "0-345-41826-3");
+        FullName fullName = new("William", "Goldman");
+        Book book = new(fullName, "The Princess Bride", "0-345-41826-3");
 
-            Assert.NotNull(book);
-            Assert.Equal("William Goldman", book.Author.ToString());
-            Assert.Equal("The Princess Bride", book.Title);
-            Assert.Equal("0-345-41826-3", book.ISBN);
-            Assert.Equal("The Princess Bride written by William Goldman. ISBN: 0-345-41826-3", book.Name);
-        }
-
-        [Fact]
-        public void Constructor_SetInvalidTitle_ThrowsArgumentNullException()
-        {
-            FullName fullName = new("William", "Goldman");
-
-            Assert.Throws<ArgumentNullException>(() => new Book(fullName, null!, "0-345-41826-3"));
-            Assert.Throws<ArgumentException>(() => new Book(fullName, "", "0-345-41826-3"));
-        }
-
-        [Fact]
-        public void Constructor_SetInvalidISBN_ThrowsArgumentNullException()
-        {
-            FullName fullName = new("William", "Goldman");
-
-            Assert.Throws<ArgumentNullException>(() => new Book(fullName, "The Princess Bride", null!));
-            Assert.Throws<ArgumentException>(() => new Book(fullName, "The Princess Bride", ""));
-        }
-
-        [Fact]
-        public void Constructor_GetsGuidAssigned_Success()
-        {
-            FullName fullName = new("William", "Goldman");
-            Book book = new(fullName, "The Princess Bride", "0-345-41826-3");
-
-            Assert.NotEqual(Guid.Empty, ((IEntity)book).Id);
-        }
-
-        [Fact]
-        public void Constructor_ComparesBooks_Success()
-        {
-            FullName fullName = new("William", "Goldman");
-            Book book1 = new(fullName, "The Princess Bride", "0-345-41826-3");
-            Book book2 = new(fullName, "The Princess Bride", "0-345-41826-3");
-            Book book3 = book1;
-
-            Assert.NotEqual(book1, book2);
-            Assert.Equal(book1, book3);
-            Assert.Same(book1, book3);
-        }   
+        Assert.NotNull(book);
+        Assert.Equal("William Goldman", book.Author.ToString());
+        Assert.Equal("The Princess Bride", book.Title);
+        Assert.Equal("0-345-41826-3", book.ISBN);
+        Assert.Equal("The Princess Bride written by William Goldman. ISBN: 0-345-41826-3", book.Name);
     }
+
+    [Fact]
+    public void Constructor_SetInvalidTitle_ThrowsArgumentNullException()
+    {
+        FullName fullName = new("William", "Goldman");
+
+        Assert.Throws<ArgumentNullException>(() => new Book(fullName, null!, "0-345-41826-3"));
+        Assert.Throws<ArgumentException>(() => new Book(fullName, "", "0-345-41826-3"));
+    }
+
+    [Fact]
+    public void Constructor_SetInvalidISBN_ThrowsArgumentNullException()
+    {
+        FullName fullName = new("William", "Goldman");
+
+        Assert.Throws<ArgumentNullException>(() => new Book(fullName, "The Princess Bride", null!));
+        Assert.Throws<ArgumentException>(() => new Book(fullName, "The Princess Bride", ""));
+    }
+
+    [Fact]
+    public void Constructor_GetsGuidAssigned_Success()
+    {
+        FullName fullName = new("William", "Goldman");
+        Book book = new(fullName, "The Princess Bride", "0-345-41826-3");
+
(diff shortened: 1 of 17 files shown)


Logger/FullName.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Logger;
+
+//FullName is a value type of record struct because we want it to be immutable.
```
LeoMarlow, line 3: I think you should provide a better explanation of why then we want it to be. 
Provide a comment on the full name record on why or why not the type is immutable. ❌

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,16 @@
...
+     *  This is an explicit impementation.
+     *  Guid ID is declared here, so that if a child wants to have a different id, ex. student id, it can be done.
+     */
+    public Guid Id { get; init; } = Guid.NewGuid();
+    /*
+     *  This is an implicit implementation.
+     *  If the child has a certain implementation for formatting on the name, it can be done in the child class.
+     *  ex. Book class
+     */
+    public abstract string Name { get; }
```
LeoMarlow, line 15: Do not implement the Name property in this abstract class. ❌

Logger/Employee.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace Logger;
+
+public record Employee : Person
+{
+    private string? _position;
+    private string? _department;
+    private string? _company;
+
+    public Employee(FullName legalName, string position, string department, string company) : base(legalName)
```
LeoMarlow, line 9: Do force any derived classes to provide an implementation for Name. ❌

Logger/IEntity.cs
```diff
@@ -1,7 +1,9 @@
-﻿namespace Logger;
+﻿
+namespace Logger;
 public interface IEntity
 {
```
LeoMarlow, line 5: Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.❌

LeoMarlow (commented):
Provide a comment on the full name record on why or why not the type is immutable. ❌
Do not implement the Name property in this abstract class. ❌
Do force any derived classes to provide an implementation for Name. ❌

Provide a comment on each interface method explaining why you implemented it implicitly or explicitly.❌
Unit test the Storage class using various entities to ensure that there are no bugs.✔
Code review (using TODO: comments if there are any issues - which there may not be) all logger-related classes.)✔

Fundamentals
Be sure you enable:

Extra Credit
Convert ILogger.CreateLogger to a generic method. ? I can't even find Ilogger
Returns the logger classes type that is to be created. ❌
Has a type parameter for the configuration type ❌

Hey guys think I found some small issues but good job over all

Logger/BaseEntity.cs
tienv9: This is not an implementation, just an abstract property and provides no default functionality.

Logger/IEntity.cs
dpalmer10: I don't think a comment here is necessary since this is the interface and nothing here is being implemented

Logger/Employee.cs
dpalmer10: Name being stored inside the constructor has been removed.

Logger/FullName.cs
dpalmer10: Probably a good idea, a better one has been written.

Logger/IEntity.cs
dpalmer10: I don't think a comment here is necessary since this is the interface and nothing here is being implemented.

1Kmiller1:
Do force any derived classes to provide an implementation for Name. ❌✔
Unit test the Storage class using various entities to ensure that there are no bugs.✔
Fundamentals
Be sure you enable:
Extra Credit
Convert ILogger.CreateLogger to a generic method. ❌
Returns the logger classes type that is to be created. ❌
Has a type parameter for the configuration type ❌


Its looking great you guys did an awesome job with the assigment and have already fixed all mistakes found by LeoMarlow, I would sugest looking into the Generic type EC it is very simple to implment. other than EC you both did a great job. though without the ILogger method that dose make it hard

Logger.Tests/BookTests.cs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    [Fact]
+    public void Constructor_GetsGuidAssigned_Success()
+    {
+        FullName fullName = new("William", "Goldman");
+        Book book = new(fullName, "The Princess Bride", "0-345-41826-3");
+        Book book2 = new(fullName, "The Princess Bride", "0-345-41826-3");
+
+        Assert.NotEqual(Guid.Empty, ((IEntity)book).Id);
```
KennethWhite, line 45: Generally we want to avoid explicit interface implementations like this, because the require casting to invoke the property.

Logger.Tests/PersonTests.cs
```diff
@@ -0,0 +1,32 @@
...
+        FullName fullName = new("Inigo", "Montoya", "A");
+        TestPerson person = new(fullName);
+        TestPerson person2 = new(fullName);
+
+        Assert.NotEqual(Guid.Empty, ((IEntity)person).Id);
+        Assert.NotEqual(((IEntity)person).Id, ((IEntity)person2).Id);
+    }
+}
+
+public record class TestPerson(FullName LegalName) : Person(LegalName)
```
KennethWhite, line 30: `TestPerson` doesn't provide any additional value if it doesn’t introduce new properties, methods, or behavior beyond what Person already provides. As written it is basically just an alias for `Person`

Logger.Tests/StorageTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using Xunit;
+
+namespace Logger.Tests;
+
+public class StorageTests
+{
+    [Fact]
+    public void Constructor_InitializedStorage_Success()
+    {
+        Storage storage = new();
```
KennethWhite, line 10: Nit: Generally tests like this don't provide much value. Since the constructor doesn't require any parameters or the object require additional configuration, we are basically testing that C# constructs objects.

KennethWhite: That being said, erring on the side of more unit tests isn't a bad thing.

Logger/BaseEntity.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Logger;
+
+public abstract record class BaseEntity : IEntity
+{
+    /*
+     *  This is an explicit impementation.
```
KennethWhite, line 6: You may wish to review explicit vs implicit and what that means for interfaces. This really comes into play in the implementing classes, not here in the interface.

Logger/Book.cs
```diff
@@ -0,0 +1,39 @@
+﻿namespace Logger;
+
+public record Book : BaseEntity
+{
+    private string? _title;
```
KennethWhite, line 5: It doesn't make sense to declare these fields as nullable since we don't actually want to allow null values to be stored.

We see this manifesting in your property getters when you needed to use the null `!` operator to assert you were not returning null. When working in a project with nullability, generally when you use the `!` operator it indicates you are not properly checking for null or that you have a nullability mismatch like we have here.

Logger/Book.cs
```diff
@@ -0,0 +1,39 @@
...
+        Author = author;
+        Title = title;
+        ISBN = isbn;
+    }
+
+    public FullName Author { get; set; }
+
+    public string Title 
+    {
+        get => _title!;
```
KennethWhite, line 19: As noted above, the null forgiveness operator should be avoided as it indicates an actual nullability problem.

Logger/Employee.cs
```diff
@@ -0,0 +1,46 @@
+﻿namespace Logger;
+
+public record Employee : Person
+{
+    private string? _position;
```
KennethWhite, line 5: Same comment regarding nullability mismatch between field and property.

Logger/FullName.cs
```diff
@@ -0,0 +1,17 @@
...
+//record struct is immutable by design and cannot be change.
```
KennethWhite, line 4: Your properties here are read-only, but it is worth noting being a record does not make them immutable unless the record class is declared with [positional syntax](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record#positional-syntax-for-property-definition) like so:
public record FullName(string FirstName, string? MiddleName, string LastName)

record structs on the other hand are always mutable to match the behavior of tuples.

Logger/Storage.cs
```diff
@@ -23,7 +23,7 @@ public bool Contains(IEntity item)
         return Entities.FirstOrDefault(entity => 
         {
             dynamic dynamicEntity = entity;
-            return dynamicEntity.Id == expectedGuid;
+            return ((IEntity)dynamicEntity).Id == expectedGuid;
```
KennethWhite, line 26: After reviewing IEntity/BaseEntity, these casts should not be required to access the Id property.

KennethWhite (commented):
First the test coverage is great, so nice work! A few mix-ups with nullability, generally having to use the `!` operator indicates you are doing something bad in a nullable context. Also a review of what an implicit vs explicit interface implementation is may be needed. 
e.g.
```interface IExample
IExample.SomeMethod
vs
SomeMethod
```

Logger.Tests/BookTests.cs
KennethWhite: After viewing the rest of the PR, I don't think the casting should be necessary at all.

PR #46, opened 2024-10-29
Assignment5: Oscar & Leo

Oscar-Ruelas:
Collab w/ @LeoMarlow


tienv9 (approved):
I think you all did well, and nothing stood out to me as needing improvement for this assignment.

Define the Node class
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌✔
Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,27 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
```
theRoice, line 4: Since you included this in Directory.Build.Props, you don't need to include them here

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,27 @@
...
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
```
theRoice, line 6: This is also included in Directory.Build.Props

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+            count++;
+        }
+
+        // Assert
+        Assert.Equal(largeCount + 1, count);
+        Assert.Equal(head, current.Next); // Circular reference to head
+    }
+
+    [Fact]
+    public void Clear_WhenClearingAlreadyClearedList_DoesNotThrowException()
```
dpalmer10, line 189: I don't really see the point of this test, especially since just below you're testing Clear an empty list.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+        Node<object> head = new ("Head");
+
+        // Act & Assert
+        Exception exception = Record.Exception(() => head.Clear());
+        Assert.Null(exception); // Ensures Clear does not throw an exception
+        Assert.Same(head, head.Next); // The head node should still point to itself
+    }
+
+    [Fact]
+    public void Clear_OnLongList_HeadSelfReferencingAndAllNodesRemoved()
```
dpalmer10, line 235: This test also seems redundant since you've already tested Clear() on a non empty list

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+        Node<int> head = new (1);
+        head.Clear(); // Clear once to start
+
+        // Act & Assert
+        Exception exception = Record.Exception(() => head.Clear());
+        Assert.Null(exception); // Ensure no exception is thrown on re-clearing
+    }
+
+    [Fact]
+    public void Clear_WithMultipleNodes_RemovedNodesPointToThemselves()
```
dpalmer10, line 201: Isn't this essentially the same test as Clear_RemovesAllNodesExceptHead()?

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+        Node<int> head = new (1);
+        head.Clear();
+
+        // Act & Assert
+        var exception = Record.Exception(() => head.Clear());
+        Assert.Null(exception);
+    }
+
+    [Fact]
+    public void Clear_SetsEachNodeToSelfReferencing()
```
dpalmer10, line 396: Isn't this a duplicate test?

GenericsHomework.Tests/NodeTests.cs
theRoice, line 201: I might be wrong here, but you clear the nodes before assigning them to their own unique variables. On line 214 when you set current to node1.Next - node1's Next value is already itself and have lost the ability to iterate through the other nodes.

To remedy this you might consider doing:

Node node2 = node1.Next;
Node node3 = node2.Next;

node1.Clear();

Assert.Same(node1, node1.Next);
Assert.Same(node2, node2.Next);
Assert.Same(node3, node3.Next);

This will ensure you have access to your nodes and can check that they are self referential.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+
+        // Act
+        bool isReadOnly = head.IsReadOnly;
+
+        // Assert
+        Assert.False(isReadOnly);
+    }
+
+    [Fact]
+    public void Clear_OnAlreadyClearedList_DoesNotThrow()
```
dpalmer10, line 384: Isn't this a duplicate test?

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
```
theRoice, line 4: I mentioned this on your other csproj for the tests project, but you can remove the target framework and nullability since you put them both in your Directory.Build.Props file!

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,414 @@
...
+
+        // Act & Assert
+        Assert.True(head.Exists(123));
+        Assert.True(head.Exists(45.67));
+        Assert.True(head.Exists("End"));
+        Assert.False(head.Exists("Nonexistent"));
+    }
+
+    [Fact]
+    public void Append_LargeNumberOfStringNodes_MaintainsCircularReferences()
```
dpalmer10, line 162: This test seems redundant.

dpalmer10 (commented):
        Define the Node class

Extra Credit

    Implement System.Collections.Generic.ICollection<T> on the Node class ✔

Fundamentals

Everything looks good, I'd check your tests and make sure you only have what's necessary. Good work!

theRoice:
Great job y'all! Very easy to follow your code and I didn't notice anything crazy. I found one test that might need fixing and some redundancy in your csproj files but those are easy fixes!

## Instructions

- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.
  - Define the `Node` class
  - The values should be homogeneous.
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ✔

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ✔

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. ✔

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Oscar-Ruelas: Yea I know they're included in the Directory.Build.Props, but I won't be removing this from the . csproj because it's the default configuration that comes with when creating a project. The main purpose of Directory.Build.Props is to override the defaults that come from the .csproj file. 

According to Microsoft, Directory.Build.props files allow settings and configurations to be defined centrally, applying them to all projects in the directory and its subdirectories. **The primary purpose is to avoid repetition and provide consistency in project configurations across solutions, letting developers extend or override settings without altering individual .csproj files.**

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Oscar-Ruelas: Same thing with this review:

Yea I know they're included in the Directory.Build.Props, but I won't be removing this from the . csproj because it's the default configuration that comes with when creating a project. The main purpose of Directory.Build.Props is to override the defaults that come from the .csproj file.

According to Microsoft, Directory.Build.props files allow settings and configurations to be defined centrally, applying them to all projects in the directory and its subdirectories. The primary purpose is to avoid repetition and provide consistency in project configurations across solutions, letting developers extend or override settings without altering individual .csproj files.

Sources: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022,

GenericsHomework/GenericsHomework.csproj
Oscar-Ruelas: As mentioned before,

Yea I know they're included in the Directory.Build.Props, but I won't be removing this from the . csproj because it's the default configuration that comes with when creating a project. The main purpose of Directory.Build.Props is to override the defaults that come from the .csproj file.

According to Microsoft, Directory.Build.props files allow settings and configurations to be defined centrally, applying them to all projects in the directory and its subdirectories. The primary purpose is to avoid repetition and provide consistency in project configurations across solutions, letting developers extend or override settings without altering individual .csproj files.

Sources: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022,

GenericsHomework.Tests/NodeTests.cs
Oscar-Ruelas: Just a test ensures that the clear method doesn't error out with an exception when clearing out an empty list

GenericsHomework.Tests/NodeTests.cs
Oscar-Ruelas: not according to the GitHub summary report by removing this test it decreases my code coverage

GenericsHomework.Tests/NodeTests.cs
Oscar-Ruelas: not according to the GitHub summary report by removing this test it decreases my code coverage

GenericsHomework.Tests/NodeTests.cs
Oscar-Ruelas: You are correct. It has been fixed

GenericsHomework.Tests/NodeTests.cs
Oscar-Ruelas: its fixed

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Oscar-Ruelas: Ok, it's not too late I'll like to fix it and I'll make sure to do this for Assignment 6 as well. But now I'm confused so what's the purpose of the props file and why are we using for the assignments? Can you further explain it to it me? From the research that I did. I found in Reddit, stack overflow, and including YouTube videos explaining that the props files is intended to override the caproj files when needed. I've also talked to professionals that work as devs in LATAM that told me the same thing. But now I'm hearing it from you that it's actually the other way around. Not arguing, I just want to actually learn how to do it right. Mind if you provide me with documentation of what the actual purpose of having a props file or at least how's it's supposed to be properly implemented.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,403 @@
+namespace GenericsHomework.Tests;
+public class NodeTests
+{
+    private static readonly int[] ExpectedArray = { 1, 2, 3 };
+    private static readonly int[] ExpectedArrayWithOffset = { 0, 0, 1, 2, 3 };
```
quattro004, line 5: It looks like these values are not shared across the tests so I would just define them in the test methods instead of more globally at the class level.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,403 @@
...
+
+
+    [Fact]
+    public void Constructor_WithValidData_InitializesWithSelfReferencingNext()
```
quattro004, line 9: I'm not sure how useful this test is as you are inherently testing the constructor in other tests. Also, if you are going to test for valid data you might want to test for invalid data as well.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,222 @@
+﻿using System.Text;
+ using System.Collections;
+
+ namespace GenericsHomework;
+ [System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", 
+     Justification = "Benjamin is allowing us to  suppress this rule for this class.")]
+ public class Node<T> : ICollection<T>
```
quattro004, line 7: Since this Node is inheriting from ICollection I would expect the name to be indicative of that fact. For example, I would expect the name to be NodeCollection or NodeList or something like that so that the consumer knows at a glance that this is not a single object.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,222 @@
...
+ {
+     public T Data { get; private set; }
+     private Node<T> _next;
+
+     public Node<T> Next
+     {
+         get => _next;
+         private set => _next = value ?? throw new ArgumentNullException(nameof(value));
```
quattro004, line 15: Would the ArgumentNullException ever be thrown? Since this is a private setter you as the author are the one setting it so it should be safe to omit that branch of the code.

GenericsHomework/Node.cs
Oscar-Ruelas: @quattro004 Like I said in the suppression Benjamín allowed for the suppression of this. Just curious did you bother to look at the assignment requirements?? I'm just doing what I'm told to do. Anything else could cause my to loose points on this assignment. Got an issue of how the class is named take the issue to who created the assignment aka your boss. Thank you and have a great weekend.

GenericsHomework/Node.cs
Oscar-Ruelas: @quattro004 yea the test passed. TDD as its finest

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,222 @@
...
+
+
+     // ICollection<T> implementation
+     public void Add(T item)   // Add and Append are the same
+     {
+         Append(item);
+     }
+
+     public bool Contains(T item) // Exists and Contains are the same
+     {
```
quattro004, line 130: These methods seem redundant, do you really need Add and Append for example? It could be confusing to a consumer of this class.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,222 @@
...
+         int index = arrayIndex;
+
+         do
+         {
+             array[index++] = current.Data;
+             current = current.Next;
+         } while (current != this);
+     }
+
+     public bool Remove(T item) // Remove and clear are the similar
```
quattro004, line 146: The comment mentions that the remove and clear methods are similar but doesn't explain why or which to use when. A consumer of this class could be confused especially if the behavior is different.

Oscar-Ruelas: @quattro004 please read carefully at the requirements of the assignment. The requirements in the assignment for the extra does not mention that we have to refactor the code. It's confusing for you be you didn't read the requirements of the homework. Plus Benjamin told us that this allowed and we can this. Please read the requirements and talk to your bosses before making any remarks. Thank you

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,403 @@
...
+        // Arrange
+        Node<object> head = new (1);
+        head.Append("Two");
+        head.Append(3.5);
+
+        // Act
+        string result = head.ToString();
+
+        // Assert
+        Assert.Equal("1 -> Two -> 3.5 -> ", result);
```
quattro004, line 108: Is there an extra result here? I'd expect it to end at 3.5 based on the arranged code.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    [Fact]
+    public void Exists_WhenListIsEmpty_ReturnsFalse()
+    {
+        // Arrange
+        Node<int> head = new (1);
+
+        // Act
+        bool exists = head.Exists(0);
```
quattro004, line 118: I wonder if this passes because it's looking for a 0 but was created with a 1? Could be a false-positive test result.

GenericsHomework/Node.cs
quattro004: Also, in the assignment instructions it says "You should not rely on any BCL generic classes for your implementation.".

quattro004 (approved):
Overall, you got the gist and did a good job implementing the code. I don't think the Node class should implement ICollection<T>.

GenericsHomework/Node.cs
BenjaminMichaelis: > @quattro004 Like I said in the suppression Benjamín allowed for the suppression of this. Just curious did you bother to look at the assignment requirements?? I'm just doing what I'm told to do. Anything else could cause my to loose points on this assignment. Got an issue of how the class is named take the issue to who created the assignment aka your boss. Thank you and have a great weekend.

Hi, @Oscar-Ruelas, it's fine to justify your implementation; but I'm concerned about the tone of your comment. Your response came across as dismissive and condescending towards the constructive feedback. Unfortunately, your response did not acknowledge or respect their effort. In a professional setting, it's not just about being right or wrong, but also about how we communicate with each other.

I want to encourage you to be curious and open-minded when discussing your work. Just because someone might not understand your implementation doesn't mean they're wrong. Instead, let's focus on learning from each other's perspectives. I think the feedback you received is valid and worth considering, and I encourage you to ask questions and seek to understand their point of view. While there might be an exception for this assignment, they are valuable insights that can help you become a better engineer.

GenericsHomework/Node.cs
BenjaminMichaelis: `Remove` and `Clear` I should not be similar in implementation. They are distinct methods and a comment saying they are similar I would agree is confusing to the consumer. I would even say that a comment that is inaccurate here causes more confusion than if there wasn't a comment at all here.

Please make sure to acknowledge and respect the effort of anyone who provides feedback to you. It is valid and worth considering.

GenericsHomework/Node.cs
BenjaminMichaelis: I see no test covering this condition showing that it is necessary; and I do not see a logical way that `value` would ever be null since it is being set privately. Unless there is a test to prove otherwise that I am not seeing; I think you are correct in your assessment @quattro004 that this case is unnecessary.

GenericsHomework/Node.cs
Oscar-Ruelas: @BenjaminMichaelis I understand that my tone might have came out dismissive and condescending, but that was not my attention to come out this way. I am willing to take all the feedback and improve upon them like the feedback that we got from @quattro004 from the unit test, but to say that suppressing a message is not allowed and that the class name should be named correctly after you had specifically told us to suppress the message just for this assignment is simply contradicting and a bit frustrating wouldn't you say?? Perhaps the assignment instructions should be much more clear like if it were to be a real project requirement from a client, would you say?

@BenjaminMichaelis after this incident I understand, and I will no longer be justifying our/my  implementation. I'll stay silent like I have always have and just accept the the feedback, Thank you and you will no longer have any issues/ complaints from me. See you next week ;)

GenericsHomework/Node.cs
Oscar-Ruelas: @BenjaminMichaelis ok now the comment that I wrote for this review was more of a joke than a repulsive comment and yes I also agree that this case is unnecessary but as you can see that the majority of the work from this assignment was done by me. I had to add nonsense tests like this one to meet with the 80 % line coverage requirement.

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Joshua-Lester3: Hey @Oscar-Ruelas, no problem - we're not deducting points for this as far as I know. I was just trying to contribute to the conversation and encourage everyone to think about this. [This Microsoft Learn post](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#import-order) as well as [this one](https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022) sheds some light on this. To paraphrase, the .props file is imported early in the build process (so it sets the default configuration) and the .targets file is imported late in the build process (so it overrides other configuration). See the top comment on this [SO post](https://stackoverflow.com/questions/63616037/using-directory-build-props-to-change-targetframeworkversion-on-netframework-pr) for a probably better paraphrase.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,222 @@
...
+         }
+
+         // Finally, set the Next node to itself to mark the list as empty.
+         // This ensures that the list now has no external references to other nodes.
+         Next = this;
+     }
+
+
+     // ICollection<T> implementation
+     public void Add(T item)   // Add and Append are the same
```
BenjaminMichaelis, line 124: nit: This would be preferable to be an explicit implementation of Add so that it is not confusing to see Append and Add on the same type

PR #47, opened 2024-10-29
Assignment5: Austin & Brady

BDullanty:
@bdullanty @austinh5544


CircluarLinkedList/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+        newNode.Next = lastNode;
+        Next = newNode;
+    }
+
+    public override string? ToString()
+    {
+        return Value?.ToString();
+    }
+
+    public bool Exists(T? expectedValue)
```
LeoMarlow, line 32: This is super clean

CircluarLinkedList/Node.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        Node<T> current = this;
+
+        do
+        {
+            if ((current.Value is null && expectedValue is null) || current.Value?.Equals(expectedValue) == true)
```
LeoMarlow, line 38: (current.Value == expectedValue) fix 

I dont believe that you need a double check here.

LeoMarlow (commented):
Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.✔
Define the Node class✔
The values should be homogeneous.✔
Extra Credit

Good job guys only one small issue with a logic but great job !

CircluarLinkedList/Node.cs
BDullanty: Thank you :)

CircluarLinkedList/Node.cs
BDullanty: When checking out why we did not implement it like you mention, I get this error:

image

Which explains why we have our implementation instead

CircluarLinkedList/CircluarLinkedList.csproj
quattro004, line 1: The instructions say: Create a class library project called "GenericsHomework.". I noticed the solution file has that name but I think that this project and its tests should as well.

quattro004 (approved):
Nice work, I like the how the usage of Theory in the tests kept the code clean.

PR #48, opened 2024-10-30
Assignment5: Kevin

kflannery2001:
Worked solo


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            this.Next = addedNode;
+            addedNode.Next = this;
+        }
+        // If there are more nodes in the list... I don't think I need this else statement
+        else
+        {
+            addedNode.Next = this.Next;
+            this.Next = addedNode;
+        }
```
Oscar-Ruelas, line 46: Wouldn't It had been better to use a while loop or a do-while inside the else statement for this part? I feel that you appending  a new node in between other nodes if the linked-list contains a large amount of nodes

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private bool Exists(T newData)
+    {
+        // Cycle through the list to check if the value already exists
+        var current = this;
```
Oscar-Ruelas, line 52: when you can remember  to specifically specify the Data Type you're using. I know to using var simplifies things but it helps other programmers know your intent.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,85 @@
...
+            temp.Next = temp;
+        }
+
+        this.Next = this;
+    }
+
+    public override string ToString()
+    {
+        return "Data: " + Data;
+    }
```
Oscar-Ruelas, line 83: I feel that the ToString method could be improved like for example using "-->" to specify the order of the nodes .

Oscar-Ruelas (commented):
Instructions

Define the Node class
Extra Credit

Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌✔
Implement System.Collections.Generic.ICollection<T> on the Node class ❌✔
Fundamentals

nullable reference types is enabled ❌✔
Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ❌✔
Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ❌✔
and enabled .NET analyzers for both projects ❌✔


Great Job Kevin 😄. You code looks great. I like how well commented your code is. Don't Forget to do the remaining part of the fundamentals.

GenericsHomework/Node.cs
kflannery2001: Good call, fixed that

GenericsHomework/Node.cs
kflannery2001: That's a good idea to provide more information about the whole linked list, thanks

GenericsHomework/Node.cs
kflannery2001: That seems to make more sense, thanks

Joshua-Lester3:
Hey Kevin, just a heads up that we require you to have a partner for these assignments. We deduct points for this.

quattro004 (approved):
Looks good

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,82 @@
...
+        
+    }
+
+    private bool Exists(T newData)
+    {
+        // Cycle through the list to check if the value already exists
+        Node<T> current = this;
+        do
+        {
+            if (current.Data!.Equals(newData))
```
BenjaminMichaelis, line 52: I don't see how you can guarantee that Data will not be null here

PR #49, opened 2024-10-30
Assignment5: Tyler + Eric

twoody0:
Partnered with @TheRoice


kflannery2001:
I will review this

twoody0:
> Create a class library project called "GenericsHomework.". ✔ Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item. Define the Node class The values should be homogeneous. Include a constructor that takes a value. (No validation is necessary on the value). ✔ Add a ToString() override that writes out the value's ToString() result. ✔ Add a Next property that references the next node or else refers back to itself if there are no other nodes in the list. ✔ The Next property should be non-nullable (careful to follow the non-nullable property guidelines) ✔ The Next property setter should be private. ✔ Add an Append method that takes a value and appends a new Node instance after the current node (by invoking the Next property). ✔ Add a Clear method that effectively removes all items from a list except the current node. Pay attention as to whether you should be concerned with the following: Whether it is sufficient to only set Next to itself ✔ Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌ _**No test showing why this is required, I am not sure how to test for that either, I think we will be going over it in class**_ Given there is a circular list of items, provide a comment to indicate whether you need to worry about garbage collection because all the items point to each other and, therefore, may never be garbage collected. ✔ Create an Exists method to test whether a value exists in the list. ✔ Throw an error on an attempt to Append a duplicate value. (Make sure you test for this case) ✔ You should not rely on any BCL generic classes for your implementation. ✔ Extra Credit Do one of the following two options (or both if you want extra, extra credit) :)
> 
> Implement a VennDiagram structure that contains n Circles that only contains homogenous reference types of any type. ❌ Each circle contains n items and each item can belong to one or more Circle instances. You are not required to use a Node from earlier in the homework for your Venn diagram implementation. You are welcome to use existing BCL generic classes for the extra credit. Implement System.Collections.Generic.ICollection on the Node class ❌
> 
> Fundamentals Place all shared project properties into a Directory.Build.Props file.✔ nullable reference types is enabled ✔ Ensure that you turn on code analysis for all projects(EnableNETAnalyzers) ✔ Set LangVersion and the TargetFramework to the latest released versions available (preview versions optional) ✔ and enabled .NET analyzers for both projects ✔ For this assignment, always use Assert.AreEqual() (the generic version) ✔ All of the above should be unit tested ✔ Choose simplicity over complexity ✔

twoody0:
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌ _**No test showing why this is required, I am not sure how to test for that either, I think we will be going over it in class**

I believe I made a test showing why it's required for removed items to circle back on themselves

ccornell88:
Define the Node class
Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required).  ✔
No test showing why this is required, I am not sure how to test for that either, I think we will be going over it in class
Extra Credit

Implement System.Collections.Generic.ICollection on the Node class ❌

Fundamentals
For this assignment, always use Assert.AreEqual() (the generic version) ✔

good work! looks like you got the test figured out for the only thing that wasnt done except the extra credit.

LinkedList/CircularLinkedList.csproj
quattro004, line 1: The instructions say to "Create a class library project called "GenericsHomework.". The solution file has the correct name but not the library or test projects.

quattro004 (approved):
Nice job

LinkedList/CircularLinkedList.csproj
twoody0: Yeah, me and Eric thought that it would make more sense for the class library and the test class to be named CircularLinkedLsit and CircularLinkedList.Tests to be more descriptive. But I'll rename the class library to "GenericsHomework" to better follow the guidelines. Thanks for catching that!

LinkedList/Node.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            return true;
+        }
+
+        while (current.Next != this)
+        {
+            current = current.Next;
+            if (current.Value != null && current.Value.Equals(value))
+            {
+                return true;
```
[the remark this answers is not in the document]
twoody0: Good catch, I refactored it to be a do while loop to have DRY code.

LinkedList/Node.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public void Append(T value)
+    {
+        if (Exists(value))
+        {
+            throw new InvalidOperationException("Value already exists in this list.");
+        }
+
+        Node<T> node = new(value);
+        Node<T> current = this;
```
BenjaminMichaelis, line 29: What is the need of current when you have access to this implicitly?

LinkedList/Node.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        while (current.Next != this)
+        {
+            current = current.Next;
+        }
```
BenjaminMichaelis, line 34: I think this appends the new node to the end of the list, while we just need to append after the current node.

ex: 
```c#
Node<T> newNode = new(value)
{
    Next = Next
};
Next = newNode;
```

PR #51, opened 2024-11-04
Assignment5: David & Kasey

dav1dmoore:
Collab with @1Kmiller1 Kasey Miller.


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,63 @@
...
+        
+
+        Node<T> newNode = new(value)
+        {
+            Next = this.Next // New Node points back to original.
+        };
+
+        this.Next = newNode; // current node now points to new node, using this. for clarity.
+    }
+
```
AustinH5544, line 33: After class, when we went to talk to Ben about the Clear() method, he told us that it would be smart to go through the list and set all the Nodes back to themselves. The reason being, if one of the nodes was saved in a variable somewhere, a portion of the list would still exist. By setting every node's Next reference to itself, we can be 100% when clear is called that any node not stored, will be garbage collected. 

As a side note, as discussed today in class, if there are no stored references to any of the nodes, then setting Next to this will work just fine. 

Everything else looks great!

AustinH5544 (commented):
Define the Node class

GenericsHomework/Node.cs
dav1dmoore: I initially had the loop to close each node on itself, however after listening to Mark I removed it. I will replace the loop as Ben suggests to ensure all cases are covered. Great idea. Thanks.

BDullanty (commented):
- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.✔
  - Define the `Node` class✔
  - The values should be homogeneous.✔
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ❌ Needs to provide a test that stores a second node that is also cleared

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,116 @@
+using Xunit;
+
+namespace GenericsHomework.Tests;
+
+public class NodeTests
+{
+    [Fact]
+    public void Node_SetValue_newNodeEqualtoValue()
```
quattro004, line 8: I think the name of this test could be improved, I had to read it a couple times to understand what was being tested.

quattro004 (approved):
Nice work. Some of the test names could be updated to be more self-describing but overall looks good.

Joshua-Lester3:
Hey @1Kmiller1, we saw that you have no commits on this PR. For these assignments, we require a sizeable contribution from each team member in order to earn credit for the assignment. Please keep this in mind for future assignments. Let us know if you have any questions.

PR #52, opened 2024-11-04
Assignment 5: Ben Gales + Jake Lucas

JacobLucas02:
collaboration with @webnoben


quattro004 (commented):
Not sure if you noticed but it looks like the build failed. https://github.com/IntelliTect-Samples/EWU-CSCD371-2024-Fall/pull/52/checks

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+	<LangVersion>latest</LangVersion>	
+	<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MSTest.TestFramework" Version="3.6.2" />
```
quattro004, line 16: You shouldn't have a test package reference in your library project.

GenericsHomeworkTests/NodeTests.cs
```diff
@@ -0,0 +1,181 @@
...
+
+    [TestMethod]
+    public void Clear_MultipleClearsInARow_Passes()
+    {
+        // Arrange
+        Node<object> newNode = new("A");
+        Node<object> expectedNode = newNode.Next;
+
+        // Act
+        newNode.Clear();
```
quattro004, line 130: If the Clear function works once, wouldn't it work multiple times? I think we can assume that and not have to write a test like this.

quattro004 (approved):
Great job, loved to see all the unit tests. Just need to get the build working.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public bool Exists(T value)
+    {
+        Node<T> current = this;
+        do
+        {
+            bool? b = current.Value?.Equals(value);
+            // b will not be null when it gets here, null checked above
+            if (b!.Value)
```
BenjaminMichaelis, line 51: I'm not sure I see where the null check is occurring. Even `bool?` indicates that it expects that `b` may be null, and the `Value?` means that if Value is null, it will return null before evaluating equals, meaning that b may in fact be null.

GenericsHomework/GenericsHomework.sln
BenjaminMichaelis, line 1: It's a bit odd to have a sln file at this level and not at the root

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        // Exception for duplicate values
+        if (Exists(value)) throw new ArgumentException(nameof(value));
+
+        Node<T> newNode = new(value)
+        {
+            Next = this.Next
+        };
+
+        this.Next = newNode;
```
BenjaminMichaelis, line 27: why is `this` used here? It is unnecessary here

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,19 @@
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+	<LangVersion>latest</LangVersion>	
+	<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+	<EnableNETAnalyzers>true</EnableNETAnalyzers>
+	<CodeAnalysisTreatWarningsAsErrors>true</CodeAnalysisTreatWarningsAsErrors>
+	<EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
```
BenjaminMichaelis, line 12: These can be refactored into `props` file

PR #54, opened 2024-11-05
Assignment5: Ben Knoll + Chris Cornell

BenKnoll3:
@benknoll3 @ccornell88


1Kmiller1:
Define the Node class

Extra Credit

Fundamentals

This Assigment has been done amazingly. you hit each and every one of the assigment reqerments. I especaly enjoy the detailed comments on each method. it made it easy to read and quick to understand each method and process.

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        // Arrange
+        Node<int> node = new(23);
+
+        // Act
+        node.Append(99);
+
+        // Assert
+        Assert.Equal(99, node.Next.Value);
+        Assert.Same(node, node.Next.Next); // should point back to node.
```
twoody0, line 30: // Assert.Same(node, node.Next.Next);

twoody0 (commented):
## Instructions

- Create a node class that can contain a value of any type and points to the next node and traversing the next node points back to the first item.
  - Define the `Node` class
  - The values should be homogeneous.
    - Whether to set the removed items to circle back on themselves. In other words, whether to close the loop of the removed items. (Provide a test to show why this is required if it is required). ✔

## Extra Credit

1. Implement `System.Collections.Generic.ICollection<T>` on the `Node` class ❌

## Fundamentals

## Overall

- You guys did a really good job! The only things I could find were nitpicky things such as the logic comments, or other small stuff.
- Just remember to leave a comment explaining whether you need to worry about garbage collection or not, unless I missed something from your comments.

GenericsHomework.Tests/NodeTests.cs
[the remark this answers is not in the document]
ccornell88: got it thanks

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                return true;
+            }
+            current = current.Next;
+        } while (current != this); // Continue until back to start
+
+        return false;
+    }
+    public void Clear()
+    {
```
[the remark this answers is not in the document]
ccornell88: got it thanks

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,93 @@
+﻿namespace GenericsHomework.Tests;
+
+public class NodeTests
+{
+    [Fact]
+    public void ToString_Returns_ProperToString()
+    {
+        // Arrange
+        Node<string> node = new("Toad");
+        string toad = "Toad";
```
quattro004, line 10: nit: you could've used this string in the line above as well to keep it DRY

quattro004 (approved):
Looks good, would love to see more tests.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+        Node<T> newNode = new(value);
+        Node<T> current = this;
+
+        while (current.Next != this)
+        {
+            current = current.Next;
+        }
+
+        current.Next = newNode;
+        newNode.Next = this;
```
BenjaminMichaelis, line 35: It looks like this appends the newNode to the end of the list instead of after the current node

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+            iterate through the entire list to ensure all Nodes are unreachable.
+        */
+        while (current != this)
+        {
+            Node<T> nextNode = current.Next;
+            previous.Next = previous;
+            current = nextNode;
+        }
+
+        this.Next = this;
```
BenjaminMichaelis, line 72:
```suggestion
        Next = this;
```

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,74 @@
...
+                return true;
+            }
+            current = current.Next;
+        } while (current != this);
+
+        return false;
+    }
+    public void Clear()
+    {
+        Node<T> current = this.Next;
```
BenjaminMichaelis, line 56:
```suggestion
        Node<T> current = Next;
```

PR #55, opened 2024-11-05
Assignment5: Eric Vo

tienv9:


GenericsHomework/Node.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            node = node.Next;
+        }
+        node.Next = Next;
+        Next = this;
+        //this set the node to point to itself, which make the list empty
+        //We don't have to worry about garbage collector because all nodes is disconnected and will be automatically collected
+    }
+
+}
```
Pbrown34, line 76: Create a class library project called "GenericsHomework.". ✔
Define the Node class
Extra Credit

Fundamentals

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            return null;
+        }
+        return _value.ToString();
+    }
+
+    public Node<T> Next
+    {
+        get { return _next; }
+        set { _next = value; }
```
quattro004, line 29: According to the instructions "The Next property setter should be private."

GenericsHomework.Tests/NodeTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        Node<int> node = new(1);
+        node.Append(2);
+        node.Append(3);
+        Assert.False(node.Remove(4));
+    }
+
+    [Fact]
+    public void CopyTo_CopyValues_Success()
+    {
+        Node<int> list = new(0); //this is necessary
```
quattro004, line 158: Consider explaining why this is necessary.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+            previous = node;
+            node = node.Next;
+        } while (node != this);
+        return false;
+    }
+
+    public void CopyTo(T[] array, int arrayIndex)
+    {
+        //THE TEST IS WRONG NOT THIS CODE
```
quattro004, line 130: Is this comment still valid? It looks like the tests are passing, maybe the test for this is passing with a false-positive?

quattro004 (approved):
Nice job

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
+﻿using System.Collections;
+
+namespace GenericsHomework;
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "Benjamin is allowing this suppression.")]
+public class Node<T> : ICollection<T>
+{
+    private readonly T _value;
+    private Node<T> _next;
```
BenjaminMichaelis, line 8: This should be a `Next` property with a private set, not a field

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+    public Node(T value)
+    {
+        _value = value;
+        _next = this;
+    }
+
+    public T Value { get {  return _value; } }
```
BenjaminMichaelis, line 15:
```suggestion
    public T Value { get; }
```
An auto property is preferred to be used here

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+
+    public override string? ToString()
+    {
+        if (_value == null)
+        {
+            return null;
+        }
+        return _value.ToString();
```
BenjaminMichaelis, line 23:
```suggestion
        return _value?.ToString();
```
This should be able to be simplified

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+        Node<T> newNode = new(value);
+        Node<T> node = this;
+
+        while (node.Next != this)
+        {
+            node = node.Next;
+        }
+
+        node.Next = newNode;
+        newNode.Next = this;
```
BenjaminMichaelis, line 64: It looks like this appends the node to the end of the list instead of after the current node

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,154 @@
...
+                count++;
+                node = node.Next;
+            } while (node != this);
+            return count;
+        }
+    }
+
+    public bool IsReadOnly => false;
+
+    public void Add(T item)
```
BenjaminMichaelis, line 101: nit: This would be preferable to be an explicit implementation of Add so that it is not confusing to see Append and Add on the same type

PR #56, opened 2024-11-05
Assignment6: Austin + Brady

AustinH5544:
@BDullanty
@AustinH5544


Calculate/Program.cs
```diff
@@ -0,0 +1,54 @@
+
+using System.Globalization;
+
+namespace Calculate;
+public class Program
+{
+    public Func<string?> ReadLine { get; init; }
```
dpalmer10, line 7: I think you just need to move lines 12 and 13 to 7 and 8 and then you have an empty default constructor
```suggestion
    public Func<string?> ReadLine { get; init; } = Console.ReadLine;
    public Action<string> WriteLine { get; init; } = Console.WriteLine;
    
    public Program() {}
```

dpalmer10 (commented):
Looks great! just 1 minor fix but other than that great work


    Create a Console project called "Calculate.". ✔
    Define a Program Class
        Define two init-only setter properties, WriteLine and ReadLine, that contain delegates for writing a line of text and reading a line of text respectively ✔
        Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ✔
        Set the default behavior for the WriteLine and ReadLine properties to invoke System.Console versions of the methods and add an empty default constructor. ❌
    Define a Calculator class ✔
        Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ✔
        Define a read-only property, MathematicalOperations, of type System.Collections.Generics.IReadOnlyDictionary<TKey,TValue> that:
            is initialized to a System.Collections.Generics.Dictionary<<TKey,TValue> instance that. ✔
                Uses char for the key corresponding to the operators +, -, *, and /. ✔
                Has values that correspond with the Add, Subtract, Multiple, and Divide methods. ✔
        Implement a TryCalculate method following "TryParse" pattern ✔
            Valid calculation expressions include such strings as "3 + 4", "42 - 2", etc. ✔
            If there is no whitespace around the operator, you can assume the calculation is invalid and return false. Similarly if the operands are not integers. ✔
            Use string.Split(), pattern matching, logical and operators to parse the string in their entirety ✔
            Index into the MathematicalOperations method using the operator parsed during pattern matching to find the corresponding implementation and invoke it. ✔
    Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
    Be sure to use the WriteLine/ReadLine properties on Program for testing the input and output of your program. ✔

Extra Credit

    Refactor the redirect portion of the Program class into 'ProgramBase`
    Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects
    Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0)

Fundamentals

    Place all shared project properties into a Directory.Build.props file.
    Place all shared project items into a Directory.Build.targets file. (optional)

Calculate/Program.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Calculate;
+public class Program
+{
+    public Func<string?> ReadLine { get; init; } = Console.ReadLine;
+    public Action<string> WriteLine { get; init; } = Console.WriteLine;
+
+    public Program() { }
+    public Program(Func<string?> readline, Action<string> writeLine)
+    {
+
```
julc14, line 13: nit: whitespace

Calculate/Calculator.cs
```diff
@@ -0,0 +1,54 @@
...
+            return false;
+        }
+
+        if (split[1].Length != 1 || !MathematicalOperations.ContainsKey(split[1][0]))
+        {
+            return false;
+        }
+
+        char givenOperator = split[1][0];
+        MathematicalOperations[givenOperator](left, right, out result);
```
julc14, line 50: This should be wrapped in a try catch. Divide can throw a DivideByZero Exception. To follow the try pattern that should be caught and return false.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,54 @@
...
+        MathematicalOperations = new Dictionary<char, MathOperation>
+            {
+                { '+', Add },
+                { '-', Subtract },
+                { '*', Multiply },
+                { '/', Divide }
+            };
+    }
+
+    public static void Add<T, T2>(T num1, T2 num2, out double result) where T : IConvertible where T2 : IConvertible => result = Convert.ToDouble(num1, new CultureInfo("en-US")) + Convert.ToDouble(num2, new CultureInfo("en-US"));
```
julc14, line 24: Return double here instead of as an Out Param

Calculate/Calculator.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    public static void Subtract<T, T2>(T num1, T2 num2, out double result) where T : IConvertible where T2 : IConvertible => result = Convert.ToDouble(num1, new CultureInfo("en-US")) - Convert.ToDouble(num2, new CultureInfo("en-US"));
+
+    public static void Multiply<T, T2>(T num1, T2 num2, out double result) where T : IConvertible where T2 : IConvertible => result = Math.Round(Convert.ToDouble(num1, new CultureInfo("en-US")) * Convert.ToDouble(num2, new CultureInfo("en-US")), 12);
```
julc14, line 28: Regarding the extra credit... It would have been preferable to use INumber. That solution would be cleaner and more extensible

Calculate/Calculator.cs
BDullanty: Here in the readme, it is requested we have an out parameter:
Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ✔

Calculate/Program.cs
BDullanty: Fixed :)

Calculate/Calculator.cs
BDullanty: Requested change implemented.

Calculate/Calculator.cs
BDullanty: I see, Try catch added for this method, and another null check was added for a second exception.

Thanks :)

Calculate/Calculator.cs
AustinH5544: I agree with Brady here, the ReadMe specifically requests that we have an out parameter.

Calculate/Calculator.cs
AustinH5544: This was a good call, thank you!

Calculate/Calculator.cs
julc14: This is a fair point. I read the 3rd "parameter" as the return value. But well that is not exactly a parameter. I will check with Ben on this

julc14: Given the readme directions, I'm going to restore points I took off for this

PR #57, opened 2024-11-06
Assignment5: Parker Brown

Pbrown34:


Joshua-Lester3:
Hey Parker, just a heads up that we require you to have a partner for these assignments. We deduct points for not having a partner.

GenericsHomework.Tests/GenericsHomework.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
```
BenKnoll3, line 4: With the dependencies centralized in the Directory.Build.props, only ones unique to each project need to be within the csproj files. (Framework / Nullable / ImplicitUsings can be removed here)

GenericsHomework/GenericsHomework.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>net8.0</TargetFramework>
```
BenKnoll3, line 4: Same as other project file

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,102 @@
+namespace GenericsHomework.Tests;
+using System;
+using Xunit;
+
+public class NodeTests
+{
+    [Fact]
+    public void AppendStringDataThrowsArgumentInvalidOperationException()
```
BenKnoll3, line 8: It is a bit hard to read your method names, consider using underscores. Also, if you removed them due to warnings, I think we can justify suppressing these warnings.

GenericsHomework.sln
```diff
@@ -0,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GenericsHomework", "GenericsHomework\GenericsHomework.csproj", "{A4FA4D3D-55BE-4E04-82E1-B0D3E8E9CA26}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "GenericsHomework.Tests", "GenericsHomework.Tests\GenericsHomework.Tests.csproj", "{5A5EF44D-68E2-4699-980C-8B9F4A78D284}"
+EndProject
```
BenKnoll3, line 9: I dont see your solution items here in the sln. I believe you may need to add a solution item folder to your project and add the Directory.Build.props file in order to get it to be recognized by the sln.

GenericsHomework/Node.cs
```diff
@@ -0,0 +1,67 @@
+namespace GenericsHomework;
+
+public class Node<T>
+{
+    public T Data { get; }
+    public Node<T> Next { get; set; }
```
BenKnoll3, line 6: The Next property setter should be private. ❌

BenKnoll3 (commented):
Define the Node class
The Next property setter should be private. ❌
Extra Credit

Fundamentals

Overall good work! There were just a few little ticky tack things that I found but most of the requirements are satisfied.

GenericsHomework.Tests/GenericsHomework.Tests.csproj
Pbrown34: Fixed

GenericsHomework/GenericsHomework.csproj
Pbrown34: Fixed

GenericsHomework.sln
Pbrown34: From the reading I did, its my understanding that you do not need to explicitly add your props file to your sln so long as its in the root folder of the project. It will be Discovered and applied by the MSBuild.

GenericsHomework/Node.cs
Pbrown34: Fixed

GenericsHomework.Tests/NodeTest.cs
Pbrown34: Fixed

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,106 @@
+namespace GenericsHomework.Tests;
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Xunit;
+
+public class NodeTests
+{
+    [SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Unit test method names")]
+    [Fact]
+    public void Append_StringData_ThrowsArgumentInvalidOperationException()
```
quattro004, line 10: Consider indicating in the test name that the data is duplicated. And I think you could omit the full exception name and just end with _Throws or _ThrowsException, the reader can see what type of exception is being thrown in the code.

GenericsHomework.Tests/NodeTest.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        // Arrange
+        Node<string> node = new("data");
+        // Act & Assert
+        Assert.Throws<InvalidOperationException>(() => node.Append("data"));
+    }
+
+    [SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Unit test method names")]
```
quattro004, line 18: I think you can apply this once at the class level and remove it from each method

quattro004 (approved):
Nice job

PR #59, opened 2024-11-06
Assignment 6: Tyler + Eric

twoody0:
Partnered with @TheRoice


kflannery2001:
I will review this

Calculate/Program.cs
```diff
@@ -0,0 +1,39 @@
+using Calculate.Core;
+using ConsoleUtilities;
+
+namespace Calculate;
+
+public class Program : ProgramBase
```
kflannery2001, line 6: - Set the default behavior for the `WriteLine` and `ReadLine` properties to invoke `System.Console` versions of the methods and add an empty default constructor. ❌
  _-**No empty default constructor**_

kflannery2001:
Other than the missing constructor, everything looks great to me, nice job!

Calculate/Program.cs
twoody0: I noticed the README says to 'add an empty default constructor.' I interpreted this as allowing the implicit default constructor provided by inheritance, as it achieves the same effect. Do you think it would it be better to explicitly define an empty constructor for clarity, even though it’s redundant?

kflannery2001: I think the README assumes that you've implemented a constructor to set the custom ReadLine and WriteLine properties, therefore pushing the need for the additional empty default constructor. I'm not sure how much it matters, given that your code achieves the same thing.

twoody0: My implementation achieves the same defaults by directly initializing the fields, simplifying the class by avoiding redundant constructors. Also, given that the README specifies a 'default constructor,' I interpreted this as not requiring an explicit one.

theRoice: Tyler and I spent a little bit of time discussing whether or not we wanted to explicitly define the constructor or settle for the default constructor. Since we were setting fields on creation we opted for the default as we couldn't think of any particular advantage to defining a constructor. Thanks for having us revisit this!

Oscar-Ruelas:
Create a Console project called "Calculate.". ✔
Define a Program Class
Define two init-only setter properties, WriteLine and ReadLine, that contain delegates for writing a line of text and reading a line of text respectively ✔
Write a test that sets these properties at construction time and then invokes the properties and verifies the expected behavior occurs. ✔
Set the default behavior for the WriteLine and ReadLine properties to invoke System.Console versions of the methods and add an empty default constructor. ✔
Define a Calculator class ✔
Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. ✔
Define a read-only property, MathematicalOperations, of type System.Collections.Generics.IReadOnlyDictionary<TKey,TValue> that:
is initialized to a System.Collections.Generics.Dictionary<<TKey,TValue> instance that. ✔
Uses char for the key corresponding to the operators +, -, *, and /. ✔
Has values that correspond with the Add, Subtract, Multiple, and Divide methods. ✔
Implement a TryCalculate method following "TryParse" pattern ✔
Valid calculation expressions include such strings as "3 + 4", "42 - 2", etc. ✔
If there is no whitespace around the operator, you can assume the calculation is invalid and return false. Similarly if the operands are not integers. ✔
Use string.Split(), pattern matching, logical and operators to parse the string in their entirety ✔
Index into the MathematicalOperations method using the operator parsed during pattern matching to find the corresponding implementation and invoke it. ✔
Implement the Program class to instantiate the calculator and invoke it based on user input from the console. ✔
Be sure to use the WriteLine/ReadLine properties on Program for testing the input and output of your program. ✔
Extra Credit

Refactor the redirect portion of the Program class into 'ProgramBase` ✔
Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects ✔
Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0) ✔
Fundamentals

Place all shared project items into a Directory.Build.targets file.

twoody0:
> Overall nice job guys, you guys did an excellent job with this assignment just a couple of this that need to be addressed, but other than that your code is well organized and structured.

Thank you Oscar! Out of curiosity, what do you think needs to be addressed? Since you didn't make any comments on our code, it would help point us in the right direction if you did!

Oscar-Ruelas:
My apologies, Tyler I forgot to hit the review button. You should now see the comments that I made on your code

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,77 @@
...
+
+    public bool Subtract(T a, T b, out double result) => (result = ConvertToDouble(a) - ConvertToDouble(b)) == result;
+
+    public bool Multiply(T a, T b, out double result) => (result = ConvertToDouble(a) * ConvertToDouble(b)) == result;
+
+    public bool Divide(T a, T b, out double result) =>
+        ConvertToDouble(b) != 0
+        ? (result = ConvertToDouble(a) / ConvertToDouble(b)) == result
+        : (result = double.NaN) == result;
+
```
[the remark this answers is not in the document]
twoody0: Yes, it's true that we could use the void return type, but since we're following the TryParse pattern, we decided to return bool.
Here is a link to Microsoft's TryParse() method: https://learn.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-8.0

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            return false;
+        }
+
+        if (!TryParse(parts[0], out T operand1) || !TryParse(parts[2], out T operand2))
+        {
+            return false;
+        }
+
+        char operatorChar = parts[1][0];
```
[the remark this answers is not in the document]
twoody0: I actually did need the [0] after [1] because it was essentially creating a string assigned to"+" in the index 1 of parts, then the [0] was retrieving the index[0] of the string just created, but because that's kind of confusing, I changed the code to assign a operatorStr to parts[1], then in the operatorChar I just accessed the 0'th index of that string, but the code is essentially the same, which is pretty interesting.

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System.Globalization;
+
+namespace Calculate.Core;
+
+public class Calculator<T> where T : struct, IConvertible
```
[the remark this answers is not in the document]
twoody0: Thank you for the recommendation! I think we're going to keep our code as is since I don't see an absolute necessary reason to refactor it to use the INumber interface, but I appreciate the recommendation, and I'll definitely keep that in mind if I have to code something similar to this project.

Calculate/Program.cs
julc14, line 6: Set the default behavior for the WriteLine and ReadLine properties to invoke System.Console versions of the methods and _add an empty default constructor_.

Missing empty constructor

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        return operation(operand1, operand2, out result);
+    }
+
+    private static double ConvertToDouble(T value)
+    {
+        return Convert.ToDouble(value, CultureInfo.InvariantCulture);
+    }
+
+    public bool Add(T a, T b, out double result) => (result = ConvertToDouble(a) + ConvertToDouble(b)) == result;
```
julc14, line 55: Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. 

Methods are not static

Calculate/Program.cs
Oscar-Ruelas: @julc14 I hope @twoody0 that I'm not intervening by answering this question. But its C# and Java 101 that the class itself creates an empty constructor for us at compile time. There's no need to for us to hardcode the empty constructor, its just redundant and looks like spaghetti code in my opinion.

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,79 @@
...
+
+    public bool Subtract(T a, T b, out double result) => (result = ConvertToDouble(a) - ConvertToDouble(b)) == result;
```
julc14, line 57: I don't see any reason why these should return a bool. In my opinion it will confuse consumers of the API. Why is multiply returning true or false... what do those results mean... Its unnecessarily complex. If you really wanted to keep it you should follow a Try pattern so the user knows something is being tried...

but even then I don't see any scenario where this would return anything other than true.

Calculate.Core/Calculator.cs
Oscar-Ruelas: @julc14 can't have static methods when using generics with> I meant that's what the build told us when it crashed with error codes, telling us to remove remove static from the methods.

Calculate.Core/Calculator.cs
julc14: Reading your comment it seems you made a design decision to simplify the internal design... everything follows an operation signature. But this comes at the cost of a more confusing external API for consumers. An add or subtract operation really shouldn't fail so wrapping it in a try is unnecessary. 

A clear and simple external API should be preferred

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,79 @@
+﻿using System.Globalization;
+
+namespace Calculate.Core;
+
+public class Calculator<T> where T : struct, IConvertible
```
julc14, line 5: Scenarios like these are exactly why Microsoft created INumber and using it should be your default unless you are able to articulate exactly why this implementation is better. 

This implementation forces all operators to go through double... but maybe I have a new number type with a unique way of dividing and I want to use the divide operator associated with this type... using INumber would enable that scenario where it is not possible here.

Calculate.Core/Calculator.cs
```diff
@@ -0,0 +1,79 @@
...
+            _mathematicalOperations.Add('*', Multiply);
+            _mathematicalOperations.Add('/', Divide);
+        }
+    }
+
+    public bool TryCalculate(string calculation, out double result)
+    {
+        result = 0;
+
+        string[]? parts = calculation.Split(' ');
```
julc14, line 28: nit: you mark this as nullable then assume the line below that is not. I believe split will never return null and if so use var or lose the question mark

ConsoleUtilities/ProgramBase.cs
```diff
@@ -0,0 +1,19 @@
+namespace ConsoleUtilities;
+
+public abstract class ProgramBase
+{
+    private Action<string> _writeLine = Console.WriteLine;
+    private Func<string> _readLine = Console.ReadLine!;
+
+    public Action<string> WriteLine
```
julc14, line 8: nit: an auto property could be used here and IMO is easier to read

Calculate.Core/Calculator.cs
julc14: You should be able to do this, which is a static call. 

Calculator<int>.Add(1,2,result)

Ill poke around after im done grading other students, but you absolutely can combine generics and static

julc14: Github formatting is making it hard to write that statement

Calculate/Program.cs
theRoice: We wrote this understanding that the class makes an empty constructor by default. I believe we even asked Mark if this was okay and he gave us the go-ahead for this assignment.

Calculate.Core/Calculator.cs
theRoice: Thanks for the insight here! We had missed INumber as an option and it would definitely make implementing a calculator class much easier.

Calculate.Core/Calculator.cs
theRoice: We did this to match TryParse. Looking at it now I believe this was a misunderstanding on our part of the line "Define static `Add`, `Subtract`, `Multiple`, and `Divide` methods that have two parameters and return a third parameter." We both took this as the parameter should take two arguments and have a third out argument. 

We intended for these to attempt an operation and return true if successful along with the result as an out. We will revisit this implementation!

Calculate/Program.cs
julc14: I am aware of that.

I'm grading based on what the requirements for the assignment dictate. If you felt like a requirement was out of place bringing that up with the instructor would have been a good option.

theRoice: That's what I meant by my last comment. We had asked the instructors if our implementation was okay due to the default nature and got the go-ahead. We should have put that in a comment for grading purposes, so I apologize for not having that annotated.

Calculate.Core/Calculator.cs
Oscar-Ruelas: Yes we are definitely treating warnings as errors. It was Mark's #1 requirement from the start of the course.

Calculate.Core/Calculator.cs
twoody0: You are right, I didn't realize that the Split method on a string always returns an array, so this is definitely good to know in the future!

ConsoleUtilities/ProgramBase.cs
twoody0: Thank you for the insight! I'll aim to prefer auto properties more in the future when backing fields aren't necessary. In this case, we decided to use backing fields because the WriteLine and ReadLine inits include additional logic that ensures null values are handled appropriately.

Calculate.Core/Calculator.cs
julc14: Given the readme directions say to use 3 params, I'm going to restore points I took off for this

theRoice: Awesome thank you! I do agree that this is unnecessarily complex for a calculator class. Maybe a more appropriate name for these particular methods in their current state would be something along the lines of tryAdd, trySubtract, tryMultiply, and tryDivide?

PR #61, opened 2024-11-07
Assignment6: Kevin

kflannery2001:
Worked solo


Calculate/Calculator.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.Runtime.Intrinsics.X86;
+using System.Text;
+using System.Threading.Tasks;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace Calculate;
+
+public class Calculator
+{
+    public static double Add(double left, double right) => left + right;
```
theRoice, line 15: I like that you used lambda statements for the add, subtract, multiply, and divide methods. This looks clean and works well for simple operations.

theRoice:
## Instructions

- Define a Program Class
  - Define two init-only setter properties, `WriteLine` and `ReadLine`, that contain delegates for writing a line of text and reading a line of text respectively ✔
  - Set the default behavior for the `WriteLine` and `ReadLine` properties to invoke `System.Console` versions of the methods and add an empty default constructor. ✔
- Define a Calculator class ✔

## Extra Credit

- Refactor the redirect portion of the `Program` class into 'ProgramBase` ❌
- Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects ❌
- Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0) ❌

## Fundamentals

- Place all shared project properties into a `Directory.Build.Props` file. ✔
- Place all shared project items into a `Directory.Build.targets` file. ✔

theRoice:
This looks great, Kevin! I wish I had more feedback for you!

Calculate/Calculator.cs
theRoice: Hey Kevin, I was prompted to revisit this portion of the code and I think you might want to revise the arithmetic methods. The way the README is written, you'll want these to functions to mimic how the TryParse library is written. The TryParse methods have boolean return values and also have a third 'out' parameter that returns the result of the calculation performed.

You can find the library [here](https://learn.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=net-8.0).
You can also view our [pull request](https://github.com/IntelliTect-Samples/EWU-CSCD371-2024-Fall/pull/59) to see how we decided to implement this on our end. 

Hope this helps!

kflannery2001: Good catch, I've refactored those methods. I appreciate the resources!

theRoice: Of course! Your code served as inspiration for us to refactor our arithmetic functions into lambda statements! I really think that lambdas are the way to go for simple operation like these but getting used to seeing where to use them is proving challenging currently.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Threading.Tasks;
+using static System.Runtime.InteropServices.JavaScript.JSType;
+
+namespace Calculate;
+
+public class Calculator
+{
+    public static bool Add(double left, double right, out double result)
+    {
+        return (result = left + right) == result;
```
julc14, line 17: I don't see any reason why these should return a bool. In my opinion it will confuse consumers of the API. Why is multiply returning true or false... what do those results mean... Its unnecessarily complex. If you really wanted to keep it you should follow a Try pattern so the user knows something is being tried...

but even then I don't see any scenario where this would return anything other than true.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,63 @@
...
+    public IReadOnlyDictionary<char, MathOperation> MathematicalOperations { get; }
+        = new Dictionary<char, MathOperation>
+    {
+            ['+'] = Add,
+            ['-'] = Subtract,
+            ['*'] = Multiply,
+            ['/'] = Divide
+    };
+
+    public bool TryCalculate(string input, out double result)
```
julc14, line 44: This method should contain a try-catch to align with the try pattern. Currently it could throw a DivideByZero Exception. It should catch this and return false in that scenario

Calculate.Tests/ProgamTests.cs
```diff
@@ -0,0 +1,31 @@
+using System.Security.Cryptography.X509Certificates;
+
+namespace Calculate.Tests;
+
+public class ProgamTests
+{
+    [Fact]
+    public void SetAndInvokeProperties_BehavesAsExpected_Success()
```
julc14, line 8: Should have 2 tests. One for the reader and one for the writer. Generally a unit test should only contain 1 assert

Calculate/Calculator.cs
julc14: Given the readme directions say to use 3 params, I'm going to restore points I took off for this

PR #64, opened 2024-11-12
Assignment6: Ben Knoll + Chris Cornell

BenKnoll3:
@Benknoll3 @ccornell88


Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    public void TryCalculate_PassedNonIntOperands_ReturnsFalse()
+    {
+        // Arrange
+        Calculator calculator = new();
+
+        // Act
+
+        // Assert
+        Assert.False(calculator.TryCalculate("d + 2", out int resString));
+        Assert.False(calculator.TryCalculate("2.2 + 2", out int resDouble));
```
BDullanty, line 51: Consider using theory, you are missing many tests for TryCalculate, such as 1 - 1 ect. This will help bring your test coverage to the expected 80% or above.

BDullanty (commented):
- Define a Program Class ✔
  - Define two init-only setter properties, `WriteLine` and `ReadLine`, that contain delegates for writing a line of text and reading a line of text respectively ✔
  - Set the default behavior for the `WriteLine` and `ReadLine` properties to invoke `System.Console` versions of the methods and add an empty default constructor. ✔
- Define a Calculator class ✔

## Extra Credit

- Refactor the redirect portion of the `Program` class into 'ProgramBase`❌
- Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects❌
- Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0)❌

## Fundamentals

Calculate.Tests/CalculatorTests.cs
ccornell88: Added a test using theory.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static int Multiply(int first, int second)
+    {
+        return first * second;
+    }
+
+    public static int Divide(int first, int second)
+    {
+        return first / second;
```
julc14, line 42: You will lose precession with certain combinations. 15 / 10 = 1. Not a great calculator if one gets that result. IMO divide should return a double and a or b must be converted to a double

Calculate/Calculator.cs
```diff
@@ -0,0 +1,60 @@
...
+        string[] sArray = s.Trim().Split(" ");
+
+        if (sArray.Length == 3 && Int32.TryParse(sArray[0], out int first) && Int32.TryParse(sArray[2], out int second) && Char.TryParse(sArray[1], out char mathOperator) && MathematicalOperations.ContainsKey(mathOperator))
+        {
+            Func<int, int, int> target = MathematicalOperations[mathOperator];
+
+            result = target(first, second);
+            return true;
+        }
+        result = -1;
```
julc14, line 56: To follow the try pattern, we would expect "legitimate tries" do not throw an exception. This method should catch the DivideByZeroException and return false

Calculate/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+    public static void Main()
+    {
+        Program program = new();
+        string? userInput;
+        Console.WriteLine("Welcome to ultimate calculator.\nEnsure format: 'int validOperator int'" +
+            "\nType EXIT to close the program.");
+        
+        do
+        {
+            userInput = Console.ReadLine();
```
julc14, line 38: You set WriteLine = Console.WriteLine but then do not use it here

Calculate/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Calculate;
+
+public class Program
+{
+    public Action<string> WriteLine { get; init; }
+    public Func<string?> ReadLine { get; init; }
+
+    private Calculator calculator;
+
+    public Program()
```
julc14, line 13: Set the default behavior for the WriteLine and ReadLine properties to invoke System.Console versions of the methods and add an empty default constructor. 

Constructor is not empty

Calculate/Calculator.cs
```diff
@@ -0,0 +1,60 @@
...
+    public static int Divide(int first, int second)
+    {
+        return first / second;
+    }
+
+    public bool TryCalculate(string s, out int result)
+    {
+        string[] sArray = s.Trim().Split(" ");
+
+        if (sArray.Length == 3 && Int32.TryParse(sArray[0], out int first) && Int32.TryParse(sArray[2], out int second) && Char.TryParse(sArray[1], out char mathOperator) && MathematicalOperations.ContainsKey(mathOperator))
```
julc14, line 49: nit: generally preferred to use int.tryParse instead of Int32

Calculate/Program.cs
BenKnoll3: Good catch, we completely overlooked that.

Calculate/Calculator.cs
ccornell88: Got it thanks!

PR #65, opened 2024-11-12
Assignment6: Eric Vo

tienv9:
Allowed to work solo by Benjamin


JacobLucas02:
## Instructions

- Define a Program Class
  - Define two init-only setter properties, `WriteLine` and `ReadLine`, that contain delegates for writing a line of text and reading a line of text respectively ❌
  - Set the default behavior for the `WriteLine` and `ReadLine` properties to invoke `System.Console` versions of the methods and add an empty default constructor. ❌
- Define a Calculator class ✔

## Extra Credit

- Refactor the redirect portion of the `Program` class into 'ProgramBase`
- Move ProgramBase into a ConsoleUtilities assembly to be used in other console-based projects
- Use generics the mathematical operations methods and consider using generic constraints (requires .NET 7.0)

## Fundamentals

- Place all shared project properties into a `Directory.Build.props` file.
- Place all shared project items into a `Directory.Build.targets` file. (optional)
- For this assignment, always use `Assert.AreEqual<T>()` (the generic version)  ❌

Calculate/Program.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+namespace Calculate;
+
+public class Program
+{
+    public Action<string> WriteLine { get; set; } = Console.WriteLine;
```
JacobLucas02, line 6: Both WriteLine and ReadLine should init-only setters

Calculate/Program.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+namespace Calculate;
+
+public class Program
+{
```
JacobLucas02, line 5: There should be an empty default constructor

Calculate.Tests/CalculatorTests.cs
```diff
@@ -0,0 +1,177 @@
...
+        // Arrange
+        int a = 10;
+        int b = 5;
+        int expected = 5;
+
+        // Act
+        int actual = Calculator.Subtract(a, b);
+
+        // Assert
+        Assert.Equal(expected, actual);
```
JacobLucas02, line 33: The unit-tests should use Assert.AreEqual<T>()

tienv9: I don't think I can use Assert.AreEqual() because it only exists in NUnit, not Xunit (which is what I'm using)

Calculate/Program.cs
tienv9: Thanks for pointing that out, it should be fixed now.

Calculate/Program.cs
tienv9: Thanks for pointing that out, it should be fixed now.

Calculate/Calculator.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Numerics;
+
+namespace Calculate;
+
+public class Calculator<T> where T : INumber<T>
+{
+    public T Add(T a, T b) => a + b;
```
julc14, line 9: Define static Add, Subtract, Multiple, and Divide methods that have two parameters and return a third parameter. 

Methods are not static

Calculate/Calculator.cs
```diff
@@ -0,0 +1,58 @@
...
+            try
+            {
+                char operationChar = parts[1][0];
+                Func<T, T, T> operation = MathematicalOperations[operationChar];
+                result = operation(operand1!, operand2!);
+                return true;
+            }
+            catch (FormatException)
+            {
+                throw new FormatException("Unable to parse input.");
```
julc14, line 51: Its not very useful to catch and the rethrow the same exception. You lose stack trace details this way.

Also things like a DivideByZeroException will not be caught breaking the try pattern.

You should catch all exceptions then return false

Calculate/Calculator.cs
tienv9: Since I turned this into a generic type to attempt the extra credit, it can't be static because static members cannot be declared on a generic type, or I will get an CA1000 error. 
https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1000

julc14: Thanks for pointing this out. This is a warning but if TreatErrorsAsWarning is on... well it will be an error. I think this is legitimate reasoning and will not take off any points

PR #69, opened 2024-11-19
Assignment7+8: Jake + Kevin

JacobLucas02:
collaborated with @kflannery2001


theRoice:
## Instructions

**Throughout, consider using the `System.Linq.Enumerable` methods `Zip`, `Count`, `Sort` and `Contains` methods for testing collections.**. (Preferably avoid using `Microsoft.VisualStudio.TestTools.UnitTesting.CollectionAssert` although that might be easier, to get a firmer grasp on additional LINQ API.)

1. Implement the `ISampleData.CsvRows` property, loading the data from the `People.csv` file and returning each line as a single string. ✔

   - Change the "Copy to" property on People.csv to "Copy if newer" so that the file is deployed along with your test project. ✔
   - Using LINQ, skip the first row in the `People.csv`. ✔
   - Be sure to appropriately handle resource (`IDisposable`) items correctly if applicable (and it may not be depending on how you implement it). ✔

2. Implement `IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()` to return a **sorted**, **unique** list of states. ✔

   - Use `ISampleData.CsvRows` for your data source. ✔
   - Don't forget the list should be unique. ✔
   - Sort the list alphabetically. ✔
   - Include a test that leverages a hardcoded list of addresses. ✔
   - Include a test that uses LINQ to verify the data is sorted correctly (do not use a hardcoded list).  ❌

3. Implement `ISampleData.GetAggregateSortedListOfStatesUsingCsvRows()` to return a `string` that contains a **unique**, comma separated list of states. ✔

   - Use `ISampleData.GetUniqueSortedListOfStatesGivenCsvRows()` for your data source. ✔
   - Consider "selecting" only the states and calling `ToArray()` to retrieve an array of all the state names. ✔
   - Given the array, consider using `string.Join` to combine the list into a single string. ✔

4. Implement the `ISampleData.People` property to return all the items in `People.csv` as `Person` objects ✔

   - Use `ISampleData.CsvRows` as the source of the data. ✔
   - Sort the list by State, City, and Zip. ✔
   - Be sure that `Person.Address` is also populated. ✔
   - Adding null validation to all the `Person` and `Address` properties is **optional**.
   - Consider using `ISampleData.CsvRows` in your test to verify your results. ✔

5. Implement `ISampleDate.FilterByEmailAddress(Predicate<string> filter)` to return a list of names where the email address matches the `filter`. ✔

   - Use `ISampleData.People` for your data source. ✔

6. Implement `ISampleData.GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people)` to return a `string` that contains a **unique**, comma-separated list of states. ✔

   - Use the `people` parameter from `ISampleData.People` property for your data source. ✔
   - At a minimum, use the `System.Linq.Enumerable.`Aggregate` LINQ method to create your result. ✔
   - Don't forget the list should be unique. ✔
   - It is recommended that, at a minimum, you use `ISampleData.GetUniqueSortedListOfStatesGivenCsvRows` to validate your result. ✔

7. Given the implementation of `Node` in Assignment5

- Implement `IEnumerable<T>` to return all the items in the "circle" of items. ✔
- Add an `IEnumberable<T> ChildItems(int maximum)` method to `Node` that returns the remaining items with a maximum number of items returned less than `maximum`.  ✔

## Extra Credit

- Implement the homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `IAsyncSampleData`). Refactor your `SampleData` and `SampleDataAsync` classes with minimal duplication. Be sure to refactor your tests to re-use a significant amount of the test code for both implementations.  ❌

## Fundamentals

Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        var people = sampleData.People;
+        // Act
+        var aggregateListOfStates = sampleData.GetAggregateListOfStatesGivenPeopleCollection(people);
+        var expectedStates = String.Join(", ", sampleData.GetUniqueSortedListOfStatesGivenCsvRows());
+        // Assert
+        Assert.AreEqual(expectedStates, aggregateListOfStates);
+    }
+}
+
+public class SampleDataForTesting : SampleData
```
theRoice, line 118: This is a really interesting way of testing your SampleData class! What made you guys decide to test it this way instead of testing the class itself?

Assignment.Tests/SampleDataTests.cs
theRoice, line 1: One of the readme requirements is to test without a hardcoded list in some instances. It might be worthwhile pulling results from the actual .csv to make sure that your DataSample class and its test file work with it.

Assignment.Tests/NodeTests.cs
theRoice, line 1: These tests look great!

Assignment/SampleData.cs
```diff
@@ -0,0 +1,81 @@
...
+        return CsvRows
+           .Select(row => row.Split(',')[6])
+           .Distinct()
+           .OrderBy(state => state)
+           .ToList();
+    }
+
+    // 3.
+    public string GetAggregateSortedListOfStatesUsingCsvRows()
+    {
```
theRoice, line 31: I really like the implementation of this method. We might revisit ours to make it more like this since yours fits the spirit of the readme more.

Assignment.Tests/SampleDataTests.cs
kflannery2001: We implemented the SampleDataForTesting class just as a to have our own, smaller data set which makes testing more simple/convenient.

Assignment.Tests/SampleDataTests.cs
kflannery2001: Thank you for these notes, we've considered them all and pushed commits accordingly.

Assignment.Tests/NodeTests.cs
kflannery2001: Thanks!

Assignment/SampleData.cs
kflannery2001: Thanks!

Assignment/SampleData.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            return File.Exists("People.csv") ? File.ReadLines("People.csv").Skip(1) : Enumerable.Empty<string>();
+        }
+    }
+
+    // 2.
+    public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+    {
+        return CsvRows
+           .Select(row => row.Split(',')[6])
```
BenjaminMichaelis, line 23: Using an enum or similar instead of just a number would bring more clarity to the fact that the number represents a column for state instead of just being arbitrary.

Assignment/SampleData.cs
```diff
@@ -0,0 +1,81 @@
...
+           .Distinct()
+           .OrderBy(state => state)
+           .ToList();
```
BenjaminMichaelis, line 26: Having a `.ToList()` at the end and then returning an `IEnumerable` is unnecessary and means that a minimum of 2 iterations will be done.

PR #79, opened 2024-12-04
Assignment9+10 multithreading: Ben Knoll + Eric Vo

BenKnoll3:
@benknoll3 @tienv9


ccornell88 (commented):
Implement PingProcess' public Task RunTaskAsync(string hostNameOrAddress) ✔
Implement PingProcess' async public Task RunAsync(string hostNameOrAddress) ✔
Complete/fix AND test async public Task RunAsync(IEnumerable hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
Implement AND test public Task RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task is also okay. NOTE: This method does NOT use Task.Run.
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌
Fundamentals
Place all shared project properties into a Directory.Build.Props file. ✔
Place all shared project items into a Directory.Build.targets file. ✔
For this assignment, consider using Assert.AreEqual() (the generic version) ✔

Good job guys! I couldnt really find anything wrong with your code besides a few  minor things i dont think are worth commenting on.

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,215 @@
...
+
+    // #2 & 3
+    async public Task<PingResult> RunAsync(
+        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        Task<PingResult> task = Task.Run(() =>
+            
+                {
+                    PingResult result = Run(hostNameOrAddress);
+                    cancellationToken.ThrowIfCancellationRequested();
```
ccornell88, line 55: Might be best to check for cancellation prior to invoking run to save resources

BenKnoll3: Good idea. We have made this addition to the code.

KennethWhite (commented):
Overall great work! Looks like you nailed all the criteria here, I left a couple comments but they are just minor cleanup/suggestions.

PR #80, opened 2024-12-08
Assignment9+10: Brady + Eric L

theRoice:
@BDullanty
@theRoice

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -91,17 +91,24 @@ public class PingProcessTests
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
-    async public Task RunAsync_UsingTpl_Success()
+    [DataRow("www.google.com")]
+    [DataRow("localhost")]
+    [DataRow("8.8.8.8")]
+    [DataRow("facebook.com")]
+    async public Task RunAsync_UsingTpl_Success(string hostName)
     {
+        // Arrange
         // DO use async/await in this test.
         PingResult result = default;
 
+        // Act
+        Task<PingResult> pingResult = Sut.RunAsync(hostName);
+        result = await pingResult;
+
+        //Assert
         // Test Sut.RunAsync("localhost");
         AssertValidPingOutput(result);
     }
-#pragma warning restore CS1998 // Remove this
-
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -140,8 +140,9 @@ public class PingProcessTests
         
         // Act
         int? lineCount = result.StdOutput?.Split(Environment.NewLine).Length;
-        
+
         // Assert
+        Assert.AreEqual(0, result.ExitCode);
         Assert.AreEqual(expectedLineCount, lineCount);
     }
 
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -1,4 +1,4 @@
-﻿using IntelliTect.TestTools;
+using IntelliTect.TestTools;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -137,12 +137,12 @@ public class PingProcessTests
 
     readonly string PingOutputLikeExpression = @"
 Pinging * with 32 bytes of data:
-Reply from ::1: time<*
-Reply from ::1: time<*
-Reply from ::1: time<*
-Reply from ::1: time<*
+Reply from *
+Reply from *
+Reply from *
+Reply from *
 
-Ping statistics for ::1:
+Ping statistics for *
     Packets: Sent = *, Received = *, Lost = 0 (0% loss),
 Approximate round trip times in milli-seconds:
     Minimum = *, Maximum = *, Average = *".Trim();
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -147,15 +147,13 @@ public class PingProcessTests
     }
 
     [TestMethod]
-#pragma warning disable CS1998 // Remove this
+
     async public Task RunLongRunningAsync_UsingTpl_Success()
     {
-        PingResult result = default;
-        // Test Sut.RunLongRunningAsync("localhost");
+        PingResult result =await Sut.RunLongRunningAsync("localhost");
         AssertValidPingOutput(result);
 
     }
-#pragma warning restore CS1998 // Remove this
 
     [TestMethod]
     public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -70,6 +70,7 @@ public class PingProcessTests
     }
 
     [TestMethod]
+
     [DataRow("www.google.com")]
     [DataRow("localhost")]
     [DataRow("8.8.8.8")]
@@ -106,20 +107,18 @@ public class PingProcessTests
 
     [TestMethod]
     [ExpectedException(typeof(AggregateException))]
-    public async void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
+    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
         // Arrange
-        // CancellationToken cancellationToken = new(true);
         CancellationTokenSource cancellationTokenSource = new();
-        
+
         // Act
         Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationTokenSource.Token);
         cancellationTokenSource.Cancel();
-        await pingResult;
-        
+        pingResult.Wait();
 
         // Assert
-        Assert.ThrowsException<AggregateException>(() => pingResult.Result);
+        //Assert.ThrowsException<AggregateException>(() => pingResult.Result);
     }
 
     [TestMethod]
@@ -128,17 +127,14 @@ public class PingProcessTests
     {
         // Use exception.Flatten()
         // Arrange
-        CancellationTokenSource cancellationTokenSource = new();
+        CancellationToken cancellationToken = new(true);
 
         // Act
-        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationTokenSource.Token);
-        cancellationTokenSource.Cancel();
-        #TODO: Finish this test
-
+        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationToken);
+        pingResult.Wait();
 
         // Assert
-        AggregateException exception = Assert.ThrowsException<AggregateException>(() => pingResult.Result);
-        //exception.Flatten();
+        Assert.ThrowsException<AggregateException>(() => pingResult.Result);
     }
 
     [TestMethod]
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -106,15 +106,11 @@ public class PingProcessTests
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
     {
         // Arrange
-        CancellationTokenSource cancellationTokenSource = new();
+        CancellationToken cancellationToken = new(true);
 
         // Act
-        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationTokenSource.Token);
-        cancellationTokenSource.Cancel();
+        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationToken);
         pingResult.Wait();
-
-        // Assert
-        //Assert.ThrowsException<AggregateException>(() => pingResult.Result);
     }
 
     [TestMethod]
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -97,17 +97,11 @@ public class PingProcessTests
     [DataRow("facebook.com")]
     async public Task RunAsync_UsingTpl_Success(string hostName)
     {
-        // Arrange
-        // DO use async/await in this test.
-        PingResult result = default;
-
-        // Act
+        // Arrange & Act
         Task<PingResult> pingResult = Sut.RunAsync(hostName);
-        result = await pingResult;
-
+        
         //Assert
-        // Test Sut.RunAsync("localhost");
-        AssertValidPingOutput(result);
+        AssertValidPingOutput(await pingResult);
     }
 
     [TestMethod]
@@ -128,10 +122,16 @@ public class PingProcessTests
     async public Task RunAsync_MultipleHostAddresses_True()
     {
         // Pseudo Code - don't trust it!!!
+
+        // Arrange
         string[] hostNames = new string[] { "localhost", "localhost", "localhost", "localhost" };
         int expectedLineCount = PingOutputLikeExpression.Split(Environment.NewLine).Length * hostNames.Length;
         PingResult result = await Sut.RunAsync(hostNames);
+        
+        // Act
         int? lineCount = result.StdOutput?.Split(Environment.NewLine).Length;
+        
+        // Assert
         Assert.AreEqual(expectedLineCount, lineCount);
     }
 
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -6,7 +6,6 @@ using System.Linq;
 using System.Threading.Tasks;
 
 namespace Assignment.Tests;
-
 [TestClass]
 public class PingProcessTests
 {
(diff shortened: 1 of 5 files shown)


Oscar-Ruelas:
## Instructions

- Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔
- Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ 
  - with a `TaskCanceledException` inner exception ✔ 
  - using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔
  - and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.

   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.
   NOTE: This method does **NOT** use `Task.Run`. ✔

## Extra Credit

## Fundamentals

- Place all shared project items into a `Directory.Build.targets` file.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+    {
+        (int exitCode, string? stdOutput) = Sut.Run("badaddress");
+        Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
+        stdOutput = WildcardPattern.NormalizeLineEndings(stdOutput!.Trim());
+        Assert.AreEqual<string?>(
+            "Ping request could not find host badaddress. Please check the name and try again.".Trim(),
+            stdOutput,
+            $"Output is unexpected: {stdOutput}");
+        Assert.AreEqual<int>(1, exitCode);
+    }
```
Oscar-Ruelas, line 47: This will need some modification since this code only works in a windows machine and not in a linux  machine feel free to look at my code

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("localhost");
```
Oscar-Ruelas, line 52:
```suggestion
        PingResult result = Sut.Run("-c 4 localhost");
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod]
+    [DataRow("www.google.com")]
```
Oscar-Ruelas, line 57: This will fail, no network connection

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+    [DataRow("localhost")]
+    [DataRow("8.8.8.8")]
+    [DataRow("facebook.com")]
```
Oscar-Ruelas, line 60: these will also fail, for the same reason as mentioned before

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+        //Act
+        Task<PingResult> task = Sut.RunTaskAsync(address);
+        task.Wait();
+        //Assert
+        Assert.AreEqual(0, task.Result.ExitCode);
+    }
+
+    [TestMethod]
+
+    [DataRow("www.google.com")]
```
Oscar-Ruelas, line 73: same thing here

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+
+    //[TestMethod]
+    //public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
+    //{
+    //    IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
+    //    System.Text.StringBuilder stringBuilder = new();
+    //    numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
+    //    int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+    //    Assert.AreNotEqual(lineCount, numbers.Count() + 1);
+    //}
```
Oscar-Ruelas, line 165: Use Try - catch and it will work:

```suggestion
  try
        {
            IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
            System.Text.StringBuilder stringBuilder = new();
            numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
            int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
            Assert.AreNotEqual(lineCount, numbers.Count() + 1);
        }
        catch(AggregateException) { 
            //Ignore Destination too short
        }```

File: Assignment.Tests/PingProcessTests.cs [hunk 7]
```diff
@@ -1,152 +1,190 @@
...
+        // Assert
+        Assert.AreEqual(0, result.ExitCode);
+        Assert.AreEqual(expectedLineCount, lineCount);
+    }
+
+    [TestMethod]
+
+    public async Task RunLongRunningAsync_UsingTpl_Success()
+    {
+        PingResult result = await Sut.RunLongRunningAsync("localhost");
```
Oscar-Ruelas, line 152:
```suggestion
        PingResult result = await Sut.RunLongRunningAsync("-c 4 localhost");
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+        Assert.ThrowsException<AggregateException>(pingResult.Wait).Flatten().Handle((exception) => throw exception);
+    }
+
+    [TestMethod]
+    public async Task RunAsync_MultipleHostAddresses_True()
+    {
+        // Pseudo Code - don't trust it!!!
+
+        // Arrange
+        string[] hostNames = ["localhost", "localhost", "localhost", "localhost"];
```
Oscar-Ruelas, line 136:
```suggestion
        string[] hostNames = ["-c 4 localhost", "-c 4 localhost", " -c 4 localhost", "-c 4 localhost"];
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+    [TestMethod]
+    [ExpectedException(typeof(TaskCanceledException))]
+    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
+    {
+        // Use exception.Flatten()
+        // Arrange
+        CancellationToken cancellationToken = new(true);
+
+        // Act
+        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationToken);
```
Oscar-Ruelas, line 124:
```suggestion
        Task<PingResult> pingResult = Sut.RunAsync("-c 4 localhost", cancellationToken);
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+
+    [TestMethod]
+    [ExpectedException(typeof(AggregateException))]
+    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
+    {
+        // Arrange
+        CancellationToken cancellationToken = new(true);
+
+        // Act
+        Task<PingResult> pingResult = Sut.RunAsync("localhost", cancellationToken);
```
Oscar-Ruelas, line 111:
```suggestion
        Task<PingResult> pingResult = Sut.RunAsync("-c 4 localhost", cancellationToken);
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+        Task<PingResult> pingResult = Sut.RunAsync(address);
+        pingResult.Wait();
+        //Assert
+        AssertValidPingOutput(pingResult.Result);
+
+    }
+
+    [TestMethod]
+    [DataRow("www.google.com")]
+    [DataRow("localhost")]
```
Oscar-Ruelas, line 92:
```suggestion
    [DataRow("-c 4 localhost")]
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+        Task<PingResult> task = Sut.RunTaskAsync(address);
+        task.Wait();
+        //Assert
+        Assert.AreEqual(0, task.Result.ExitCode);
+    }
+
+    [TestMethod]
+
+    [DataRow("www.google.com")]
+    [DataRow("localhost")]
```
Oscar-Ruelas, line 74:
```suggestion
    [DataRow("-c 4 localhost")]
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("localhost");
+        AssertValidPingOutput(result);
+    }
+
+    [TestMethod]
+    [DataRow("www.google.com")]
+    [DataRow("localhost")]
```
Oscar-Ruelas, line 58:
```suggestion
    [DataRow("-c 4 localhost")]
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,176 @@
...
+
+public record struct PingResult(int ExitCode, string? StdOutput);
+
+public class PingProcess
+{
+    private ProcessStartInfo StartInfo { get; } = new("ping");
+
+    public PingResult Run(string hostNameOrAddress)
+    {
+        string arguments = "-n 4 " + hostNameOrAddress;
```
Oscar-Ruelas, line 19: This windows command slightly modify to work on a linux computer and you'll be fine. Since your limiting the amount of ping's within this method, I think you can ignore all my comments in the test file where I mentioned to add "-c 4 "

Oscar-Ruelas (commented):
Hey guys nice job, code just needs to be modified to work in a linux OS right now you have set up to work in a windows OS. Oh you'll see this in lots of my code suggestions but pinging to an external domain won't work since the ubuntu server that Github actions utilizes isn't connected to the network. I learned that the hard way.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,190 @@
...
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        Sut = new();
+    }
+
+    [TestMethod]
+    public void Start_PingProcess_Success()
+    {
+        Process process = Process.Start("ping", "localhost");
```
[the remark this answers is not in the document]
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: Good catch, we have decided to stick to codes rather than the string output. Thanks!

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
theRoice: Thanks Oscar, we're working on getting these out of the tests!

BDullanty: Very true, strange it won't let us ping, thanks for the pointer!

Assignment.Tests/PingProcessTests.cs
BDullanty: That's true, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: All external pings removed :)

Assignment.Tests/PingProcessTests.cs
BDullanty: That's true, for now we will leave it commented out, but catching it is a good idea as well!

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment.Tests/PingProcessTests.cs
BDullanty: Good idea, we have made this change throughout to limit the ping number, thanks :)

Assignment/PingProcess.cs
BDullanty: We tried changing it here, but ended up changing it throughout the tests instead.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,223 @@
...
+
+
+    //External pings seem to not work in github
+    //[TestMethod]
+
+    //public void Run_GoogleDotCom_Success()
+    //{
+    //    int exitCode = Sut.Run("google.com -c 4").ExitCode;
+    //    Assert.AreEqual<int>(0, exitCode);
+    //}
```
KennethWhite, line 39: Generally we want to remove commented code, it helps improve readability and maintainability.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,223 @@
...
+
+    }
+
+    [TestMethod]
+    [DataRow("localhost -c 4")]
+    public async Task RunAsync_UsingTpl_Success(string hostName)
+    {
+        // Arrange & Act
+        Task<PingResult> taskResult = Sut.RunAsync(hostName);
+        PingResult pingResult = await taskResult;
```
KennethWhite, line 99: Nit: We can simplify this. Generally in async await, we await the calls inline
```suggestion
        PingResult pingResult = await Sut.RunAsync(hostName);
```

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,174 @@
...
+    public Task<PingResult> RunTaskAsync(string hostNameOrAddress)
+    {
+        return Task.Run(() => Run(hostNameOrAddress));
+    }
+
+    public async Task<PingResult> RunAsync(
+        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        //This feels like extra code, doesn't Task.Run check cancelation token right away and then begin polling IsCancellationRequested after?
+        cancellationToken.ThrowIfCancellationRequested();
```
KennethWhite, line 40: You are correct that the `Task.Run` tasks will check the cancellation tokens as well. The reason we do this ourselves before calling `Task.Run()` is, in the case of the `cancellationToken` already being cancelled, we prevent the creation of the Task entirely and the overhead that is associated with it.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,223 @@
...
+        // Assert
+        Assert.AreEqual(0, result.ExitCode);
+        try
+        {
+            Assert.AreEqual(windowsExpectedLineCount, lineCount);
+        }
+        catch
+        {
+            Assert.AreEqual(linuxExpectedLineCount, lineCount);
+        }
```
KennethWhite, line 158: This likely isn't something that was covered, but rather than a blind try/catch we can do a conditional check for the OS type:

```cs
    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
    {
        Assert.AreEqual(windowsExpectedLineCount, lineCount, "Line count does not match expected value for Windows.");
    }
    else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
    {
        Assert.AreEqual(linuxExpectedLineCount, lineCount, "Line count does not match expected value for Linux.");
    }
    else
    {
        Assert.Fail("Unsupported operating system.");
    }
    ```

File: Assignment.Tests/PingProcessTests.cs [hunk 20]
```diff
@@ -0,0 +1,223 @@
...
+
+    //[TestMethod]
+    //public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
+    //{
+    //    IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
+    //    System.Text.StringBuilder stringBuilder = new();
+    //    numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
+    //    int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+    //    Assert.AreNotEqual(lineCount, numbers.Count() + 1);
+    //}
```
KennethWhite, line 178: Same comment as before, we should remove commented code before submitting a PR.

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,174 @@
...
+        return new PingResult(ExitCode: tasks.Max(x => x.Result.ExitCode), StdOutput: string.Join(Environment.NewLine, tasks.Select(outputStringResult => outputStringResult.Result.StdOutput?.Trim() ?? "")));
+
+    }
+    public Task<PingResult> RunLongRunningAsync(
+        string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        return Task.Factory.StartNew(
+            () =>
+        {
+            Task<PingResult> result = RunAsync(hostNameOrAddress);
```
KennethWhite, line 60: Generally if we have a cancellation token, and the method we are calling has an overload with cancellation support, we should use that overload.
```suggestion
            Task<PingResult> result = RunAsync(hostNameOrAddress, cancellationToken );
```

KennethWhite (commented):
Overall great work! Only a few small comments on code cleanup/styling, but it looks like you nailed all the criteria.

Assignment/PingProcess.cs
BDullanty: That makes sense! Thanks for the response :)

PR #82, opened 2024-12-10
Assignment9+10: Jake + Kevin

JacobLucas02:
collaborated with @kflannery2001

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -74,6 +74,7 @@ public class PingProcessTests
         // Act
         Task<PingResult> task = Sut.RunAsync("localhost");
         task.Wait();
+
         PingResult result = task.Result;
 
         // Assert
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -57,11 +57,8 @@ public class PingProcessTests
     [TestMethod]
     public void RunTaskAsync_Success()
     {
-        // Arrange
-        var pingProcess = new PingProcess();
-
         // Act
-        Task<PingResult> task = pingProcess.RunTaskAsync("localhost");
+        Task<PingResult> task = Sut.RunTaskAsync("localhost");
         task.Wait();
 
         PingResult result = task.Result;
@@ -74,11 +71,8 @@ public class PingProcessTests
     [TestMethod]
     public void RunAsync_UsingTaskReturn_Success()
     {
-        // Arrange
-        var pingProcess = new PingProcess();
-
         // Act
-        Task<PingResult> task = pingProcess.RunAsync("localhost");
+        Task<PingResult> task = Sut.RunAsync("localhost");
         task.Wait();
         PingResult result = task.Result;
 
@@ -90,11 +84,8 @@ public class PingProcessTests
     [TestMethod]
     public async Task RunAsync_UsingTpl_Success()
     {
-        // Arrange
-        var pingProcess = new PingProcess();
-
         // Act
-        PingResult result = await pingProcess.RunAsync("localhost");
+        PingResult result = await Sut.RunAsync("localhost");
 
         // Assert
         Assert.IsTrue(result.StdOutput?.Contains("localhost") == true || result.StdOutput?.Contains("Reply from") == true);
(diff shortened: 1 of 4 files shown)


Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,182 @@
...
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        Sut = new();
+    }
+
+    [TestMethod]
+    public void Start_PingProcess_Success()
+    {
+        Process process = Process.Start("ping", "localhost");
```
dpalmer10, line 25: if you change localhost to "-c 4 localhost" you should be able to get the tests to work on the bot
```suggestion
        Process process = Process.Start("ping", "-c 4 localhost");
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,182 @@
...
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+
+    [TestMethod]
+    public void Run_GoogleDotCom_Success()
+    {
+        int exitCode = Sut.Run("google.com").ExitCode;
```
dpalmer10, line 33: since the bot doesn't have internet, if you change this to googles IP address instead of the website, the test will run
```suggestion
        int exitCode = Sut.Run("-c 8.8.8.8").ExitCode;
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,182 @@
...
+        Assert.AreEqual<int>(0, exitCode);
```
dpalmer10, line 34: this line would have to be updated as well since the bot doesn't have internet access
```suggestion
        Assert.AreEqual<int>(1, exitCode);
```

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,182 @@
...
+    public void StringBuilderAppendLine_InParallel_IsNotThreadSafe()
+    {
+        IEnumerable<int> numbers = Enumerable.Range(0, short.MaxValue);
+        System.Text.StringBuilder stringBuilder = new();
+        numbers.AsParallel().ForAll(item => stringBuilder.AppendLine(""));
+        int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+        Assert.AreNotEqual(lineCount, numbers.Count()+1);
+    }
+
+    readonly string PingOutputLikeExpression = @"
```
dpalmer10, line 161: this needs to be made private to get the bot to run your tests

dpalmer10 (commented):
    Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ❌✔ NOTE:
        StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌✔
    Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run.

Extra Credit

    Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌✔

Fundamentals

    Place all shared project items into a Directory.Build.targets file.

Seems like you guys did the " async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default)" without using IEnumerable, I'm not entirely sure if that's allowed so I don't know if I should mark that as correct or not. Other than that, seems like you guys need to do some modifying to tests, and then chase down all the errors to get the bot to run your program

Assignment.Tests/PingProcessTests.cs
kflannery2001: thanks!

Assignment.Tests/PingProcessTests.cs
kflannery2001: thanks!

kflannery2001:
Awesome, thank you for the input!

KennethWhite (commented):
It looks like you have some build issues, but other than that you checked off all the criteria!

PR #83, opened 2024-12-10
Assignment9+10: Parker Brown+ Leo Marlow

Pbrown34:
My partner for this assignment was @LeoMarlow

### DIFF
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -39,7 +39,7 @@ public class PingProcessTests
     {
         (int exitCode, string? stdOutput) = Sut.Run("badaddress");
         Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
-        stdOutput = WildCardPattern.NormalizeLineEndings(stdOutput!.Trim());
+        stdOutput = WildcardPattern.NormalizeLineEndings(stdOutput!.Trim());
         Assert.AreEqual<string?>(
             "Ping request could not find host badaddress. Please check the name and try again.".Trim(),
             stdOutput,
@@ -94,7 +94,7 @@ async public Task RunAsync_UsingTpl_Success()
     [ExpectedException(typeof(TaskCanceledException))]
     public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException()
     {
-        using (var cts = new CancellationTokenSource())
+        using (var cts = new System.Threading.CancellationTokenSource())
         {
             cts.Cancel();
             Task<PingResult> task = Sut.RunAsync("localhost", cts.Token);
@@ -113,16 +113,18 @@ async public Task RunAsync_UsingTpl_Success()
     {
         string[] hostNames = new string[] { "localhost", "localhost", "localhost", "localhost" };
         int expectedLineCount = PingOutputLikeExpression.Split(Environment.NewLine).Length * hostNames.Length;
-        PingResult result = await Sut.RunAsync(hostNames);
-        int? lineCount = result.StdOutput?.Split(Environment.NewLine).Length;
+        List<Task<PingResult>> tasks = hostNames.Select(host => Sut.RunAsync(host)).ToList();
+        PingResult[] results = await Task.WhenAll(tasks);
+        int lineCount = results.Sum(result => result.StdOutput?.Split(Environment.NewLine).Length ?? 0);
         Assert.AreEqual(expectedLineCount, lineCount);
     }
 
     [TestMethod]
     async public Task RunLongRunningAsync_UsingTpl_Success()
     {
-        // Test Sut.RunLongRunningAsync("localhost");
-        PingResult result = await Sut.RunLongRunningAsync("localhost");
+        // Test Sut.RunLongRunningAsync("localhost");  
+        ProcessStartInfo startInfo = new ProcessStartInfo("ping", "localhost");
+        PingResult result = await Sut.RunLongRunningAsync(startInfo, null, null, default);
         AssertValidPingOutput(result);
     }
     [TestMethod]
@@ -149,7 +151,7 @@ Approximate round trip times in milli-seconds:
     private void AssertValidPingOutput(int exitCode, string? stdOutput)
     {
         Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
-        stdOutput = WildCardPattern.NormalizeLineEndings(stdOutput!.Trim());
+        stdOutput = WildcardPattern.NormalizeLineEndings(stdOutput!.Trim());
         Assert.IsTrue(stdOutput?.Contains(PingOutputLikeExpression)??false,
             $"Output is unexpected: {stdOutput}");
         Assert.AreEqual<int>(0, exitCode);
--- a/Assignment/Assignment.Tests/PingProcessTests.cs
+++ b/Assignment/Assignment.Tests/PingProcessTests.cs
@@ -39,7 +39,7 @@ public class PingProcessTests
     {
         (int exitCode, string? stdOutput) = Sut.Run("badaddress");
         Assert.IsFalse(string.IsNullOrWhiteSpace(stdOutput));
-        stdOutput = WildcardPattern.NormalizeLineEndings(stdOutput!.Trim());
+        stdOutput = WildCardPattern.NormalizeLineEndings(stdOutput!.Trim());
         Assert.AreEqual<string?>(
             "Ping request could not find host badaddress. Please check the name and try again.".Trim(),
             stdOutput,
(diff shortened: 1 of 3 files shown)


Assignment/PingProcess.cs
```diff
@@ -21,46 +21,58 @@ public PingResult Run(string hostNameOrAddress)
...
-        throw new NotImplementedException();
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, default, default); 
+        string stdOutput = stringBuilder?.ToString() ?? string.Empty; 
+        if (string.IsNullOrWhiteSpace(stdOutput)) 
+        { 
+            stdOutput = $"Ping request could not find host {hostNameOrAddress}. Please check the name and try again."; 
+        } 
+        return new PingResult(process.ExitCode, stdOutput); 
+    } 
+    public Task<PingResult> RunTaskAsync(string hostNameOrAddress, CancellationToken cancellationToken = default)
```
twoody0, line 32: You shouldn't need the second parameter `CancellationToken cancellationToken = default`

Assignment/PingProcess.cs
```diff
@@ -21,46 +21,58 @@ public PingResult Run(string hostNameOrAddress)
...
-        Task task = null!;
-        await task;
-        throw new NotImplementedException();
+        StringBuilder? stringBuilder = null;
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+        Process process = RunProcessInternal(startInfo, updateStdOutput, progressError, token);
+
+        await Task.Run(() => process.WaitForExit(), token);
```
twoody0, line 73: The instructions say to use `Task.Factory.StartNew()` and use `TaskCreationOptions.LongRunning`, so just make sure to replace `Task.Run` with `Task.Factory.StartNew()`.

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,210 @@
...
+        Task<PingResult> task = Sut.RunAsync("-c 4 localhost");
+        task.Wait();
+        PingResult result = task.Result;
+        Assert.AreEqual(0, result.ExitCode);
+        Assert.IsFalse(string.IsNullOrWhiteSpace(result.StdOutput));
+        //AssertValidPingOutput(result);
+    }
+
+[TestMethod]
+async public Task RunAsync_UsingTpl_Success()
```
twoody0, line 86: Nit: White space, just move the test right one tab

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                stringBuilder.AppendLine("");
+            }
+        });
+
+        int lineCount = stringBuilder.ToString().Split(Environment.NewLine).Length;
+        Assert.AreEqual(numbers.Count() + 1, lineCount);
+    }
+
+//     private readonly string PingOutputLikeExpression = @"
```
twoody0, line 186: It seems like you completely removed the WildCardPattern class, I think if you add it back you shouldn't have to comment out this method, and the following methods, but either way I don't blame you because of how many issues it caused me and my partner. But you will need this to verify `"StdOutput must have all the ping output returned (no lines can be missing) even though intermingled."` If we get it working I'll let you know.

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,210 @@
...
+            Task<PingResult> task = Sut.RunAsync("-c 4 localhost", cts.Token);
+            try
+            {
+                task.Wait();
+            }
+            catch (AggregateException ex)
+            {
+                if (ex.InnerException is TaskCanceledException)
+                {
+                    throw new AggregateException(ex.InnerException);
```
twoody0, line 112: You shouldn't need to throw a new `AggregateException(ex.InnerException)` here when you're already expecting the exception.

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                throw ex.Flatten().InnerException ?? ex;
+            }
+        }
+    }
+    [TestMethod]
+    async public Task RunAsync_MultipleHostAddresses_True()
+    {
+        string[] hostNames = new string[] { "-c 4 localhost", "-c 4 localhost", "-c 4 localhost", "-c 4 localhost" };
+        foreach (var hostName in hostNames)
```
twoody0, line 140: I believe that since this is ran inside of the foreach loop, it negates the parallel execution since it processes each hostname one by one. I would suggest just getting rid of the foreach loop.

twoody0 (commented):
## Instructions

   Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔
   Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ with a `TaskCanceledException` inner exception ✔ using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.
   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay. ✔
   NOTE: This method does **NOT** use `Task.Run`. ✔

## Extra Credit

## Fundamentals

- Place all shared project items into a `Directory.Build.targets` file.

## Overall
You guys did a good job! What I would suggest working on is ensuring that you aren't using `Task.Run` since it explicitly states not to. Then I would try to get the standard output working. I'll update my PR and change the ❌'s to ✔'s when you fix it :)

Assignment.tests/PingProcessTests.cs
Pbrown34: fixed

Assignment/PingProcess.cs
Pbrown34: Fixed

Assignment.tests/PingProcessTests.cs
Pbrown34: Fixed.

Assignment.tests/PingProcessTests.cs
twoody0: Hey, we got it working, this is how we formatted it:
```csharp
private readonly string PingOutputLikeExpression = @"
PING * 56 data bytes
64 bytes from * (::1): icmp_seq=* ttl=* time=* ms
64 bytes from * (::1): icmp_seq=* ttl=* time=* ms
64 bytes from * (::1): icmp_seq=* ttl=* time=* ms
64 bytes from * (::1): icmp_seq=* ttl=* time=* ms

--- * ping statistics ---
* packets transmitted, * received, *% packet loss, time *ms
rtt min/avg/max/mdev = */*/*/* ms
".Trim();
```
    
I don't think there's any other way to format it than this. Just be sure to add the WildCardPattern class back and then you should be able to use AssertValidPingOutput in the tests.

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,202 @@
...
+        Assert.AreEqual<int>(2, exitCode);
+    }
+
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("-c 4 localhost");
+        Assert.AreEqual(0, result.ExitCode);
+
+      //  AssertValidPingOutput(result);
```
KennethWhite, line 58: Nit: We want to remove unused/commented code prior to opening a PR.

Assignment.tests/PingProcessTests.cs
```diff
@@ -0,0 +1,202 @@
...
+            {
+                task.Wait();
+            }
+            catch (AggregateException ex)
+            {
+                if (ex.InnerException is TaskCanceledException)
+                {
+                    throw new AggregateException(ex.InnerException);
+                }
+                throw ex.Flatten().InnerException ?? ex;
```
KennethWhite, line 117: In addition to Tyler's comment above, when you throw the InnerException here (or `ex`) we will actually lose the original stack trace on the exception. We want this information to potentially debug issues and understand where the exception was first thrown, even though this is a unit test.

Assignment/PingProcess.cs
```diff
@@ -21,47 +21,45 @@ public PingResult Run(string hostNameOrAddress)
...
-        await task;
-        throw new NotImplementedException();
+        return await Task.Run(() =>
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            return Run(hostNameOrAddress);
+        }, cancellationToken);
     }
 
     async public Task<PingResult> RunAsync(params string[] hostNameOrAddresses)
```
KennethWhite, line 47: This modifier declaration order is inconsistent from the standard. It should be:
```suggestion
    public async Task<PingResult> RunAsync(params string[] hostNameOrAddresses)

```

Assignment/PingProcess.cs
LeoMarlow: Removed

Assignment.tests/PingProcessTests.cs
LeoMarlow: I believe parker fixed this

Assignment.tests/PingProcessTests.cs
LeoMarlow: This is a wrapper for aggregate expression.  The expected was a taskcanceledExemption which then is rewrapped

Assignment/PingProcess.cs
```diff
@@ -21,47 +21,45 @@ public PingResult Run(string hostNameOrAddress)
...
-            // ...
-
-            await task.WaitAsync(default(CancellationToken));
-            return task.Result.ExitCode;
-        });
-
-        await Task.WhenAll(all);
-        int total = all.Aggregate(0, (total, item) => total + item.Result);
-        return new PingResult(total, stringBuilder?.ToString());
+        Task<PingResult>[] tasks = hostNameOrAddresses.Select((string x) => RunAsync(x)).ToArray();
```
KennethWhite, line 49: It looks like you are failing to capture all the ping output from the processes here.

KennethWhite (commented):
Overall nice work! Just a few minor cleanup issues but looks like you nailed the assignment criteria.

PR #87, opened 2024-12-11
Assignment9+10: Tyler + Austin

twoody0:
Partnered with @AustinH5544


Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,268 @@
...
+        Task task = Task.Run(() =>
+        {
+            Task<PingResult> result = Sut.RunAsync(hostNames, cts.Token);
+            cts.Cancel();
+            return result;
+        });
+
+        task.Wait();
+    }
+
```
LeoMarlow, line 129: An expected Exception of type CancelationException should be added here

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,201 @@
...
+
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, null, cancellationToken);
+
+        await Task.Run(() =>
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            process.WaitForExit();
```
LeoMarlow, line 46: I don't believe you need 2 cancellationToken.ThrowIfCancellationRequested(); Line 45 and 47

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,268 @@
...
+        // Act
+
+        // Assert
+        //AssertValidPingOutput(result);
+        Assert.AreEqual(0, result.ExitCode);
+        Assert.IsFalse(string.IsNullOrWhiteSpace(result.StdOutput));
+    }
+
+    [TestMethod]
+    [ExpectedException(typeof(AggregateException))]
```
LeoMarlow, line 114: Im not sure how you made this work. I had to flatten and rethrow as AggregateException here. When you cancel it should return a cancelationException

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        try
+        {
+            task.Wait();
+            Assert.Fail("Expected TaskCanceledException but no exception was thrown.");
+        }
+        catch (AggregateException ex)
+        {
+            AggregateException flattened = ex.Flatten();
+            Assert.IsTrue(flattened.InnerExceptions.Any(e => e is TaskCanceledException),
```
LeoMarlow, line 150: I think this may be backwards CanceledExemption should be the default and Aggregate should be the one you flatten. It works like this but is kinda confusing

LeoMarlow (commented):
Really nice work boys! I only found one issue thats sort of a non issue ? I think that your returning the wrong type of exemption with the cancelation. It seems to work tho anyway.
Anyway any extra detail is left in comments -L

Add support for an optional cancellation token to the PingProcess.RunAsync() signature. ✔
Inside the PingProcess.RunAsync() invoke the token's ThrowIfCancellationRequested() method so an exception is thrown. ✔
Test that, when cancelled from the test method, the exception thrown is an AggregateException ✔/ ?
with a TaskCanceledException inner exception ✔/ ?
using the test methods RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping ✔
and RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException ✔ respectively.
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔
NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay.
NOTE: This method does NOT use Task.Run. ✔
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌
Fundamentals
Place all shared project items into a Directory.Build.targets file.✔

Assignment.Tests/PingProcessTests.cs
BDullanty: As far as I am aware, any errors caused within a task results in an aggregate exception

Assignment.Tests/PingProcessTests.cs
BDullanty: I don't think that is needed as they do an assert that checks for this type of exception within the method, which is the same

Assignment.Tests/PingProcessTests.cs
BDullanty: Here they catch an aggregate exception and flatten it.  A flattened aggregate is still an aggregate exception. They then call the inner exceptions and check if it is a TaskCanceled Exception, which should be fine.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,253 @@
...
+                          "Expected TaskCanceledException in AggregateException.");
+        }
+        catch (TaskCanceledException)
+        {
+            Assert.IsTrue(true);
+        }
+    }
+
+    public PingProcess GetSut()
```
BDullanty, line 183: Perhaps just make Sut public if you need to return it outside (which I don't think you do), rather than a method.

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,201 @@
...
+
+    public async Task<PingResult> RunAsync(string hostNameOrAddress, CancellationToken cancellationToken = default)
+    {
+        StartInfo.Arguments = hostNameOrAddress;
+        StringBuilder? stringBuilder = null;
+
+        void updateStdOutput(string? line) =>
+            (stringBuilder ??= new StringBuilder()).AppendLine(line);
+
+        Process process = RunProcessInternal(StartInfo, updateStdOutput, null, cancellationToken);
```
BDullanty, line 41: Perhaps add this within the Task.Run itself

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,253 @@
...
+
+    [TestMethod]
+    public void RunAsync_UsingTaskReturn_Success()
+    {
+        // Arrange
+        PingResult result = default;
+
+        // Act
+        Task task = Sut.RunAsync("localhost -c 4").ContinueWith(t => result = t.Result);
+        task.Wait();
```
BDullanty, line 91: Rather than continue with, perhaps using task.Wait() then just referring to task.Result may be easier to read

Assignment.Tests/PingProcessTests.cs
```diff
@@ -0,0 +1,253 @@
...
+    {
+        // Arrange
+        string[] hostNames = ["localhost -c 4", "localhost -c 4", "localhost -c 4", "localhost -c 4"]; // Correct array syntax
+
+        // Act
+        PingResult result = await Sut.RunAsync(hostNames);
+
+        // Assert
+        Assert.IsFalse(string.IsNullOrWhiteSpace(result.StdOutput));
+        Assert.AreEqual(0, result.ExitCode);
```
BDullanty, line 199: - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
      
      Perhaps set a static number (like 9 * 4) based on Linux and check for that amount of lines

BDullanty (commented):
- Inside the `PingProcess.RunAsync()` invoke the token's `ThrowIfCancellationRequested()` method so an exception is thrown. ✔
- Test that, when cancelled from the test method, the exception thrown is an `AggregateException` ✔ 
  - with a `TaskCanceledException` inner exception ✔ 
  - using the test methods `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping` ✔
  - and `RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrappingTaskCanceledException` ✔ respectively.

   NOTE:
      - StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ❌
5. Implement **AND test** `public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token)` using `Task.Factory.StartNew()` and invoking `RunProcessInternal` with a `TaskCreation` value of `TaskCreationOptions.LongRunning` and a `TaskScheduler` value of `TaskScheduler.Current`. Returning a `Task<PingResult>` is also okay.✔
   NOTE: This method does **NOT** use `Task.Run`.✔

## Extra Credit

## Fundamentals

Good work :)

Assignment.Tests/PingProcessTests.cs
AustinH5544: We are returning them, and originally we did count the lines. But it was different on their Linux build so we decided not to count the lines. I suppose I could just check to make sure there is more than like 20 and that would be pretty solid proof but idk.

Assignment.Tests/PingProcessTests.cs
AustinH5544: I don't even think we are using this, I believe it was meant to be deleted but I will double check.

Assignment/PingProcess.cs
AustinH5544: What are your thoughts on why it should be inside?

Assignment.Tests/PingProcessTests.cs
AustinH5544: I will look into this. In your opinion would it just be easier or is it functionally better?

Assignment/PingProcess.cs
twoody0: I removed the first cancellationToken.ThrowIfCancellationRequested(); and it seems to be working fine, good catch :)

Assignment.Tests/PingProcessTests.cs
twoody0: I changed that, I do feel like it's slightly more readable

KennethWhite (commented):
This looks great! Looks like you knocked out all the assignment criteria, great code coverage as well! Also nice work on testing for the `TaskCanceledException` within the `AggregateException` without trying to rethrow.

PR #88, opened 2024-12-11
Assignment9+10 kasey miller

1Kmiller1:
Need help with step 5


Assignment/PingProcess.cs
```diff
@@ -1,173 +1,197 @@
...
+        var objectlock = new object();
+
+        try
+        {
+            var task = hostNameOrAddresses.Select(async hostOrAddress =>
+            {
+
+                cancellationToken.ThrowIfCancellationRequested();
+                PingResult result = await RunAsync(hostOrAddress, cancellationToken);
+
```
AustinH5544, line 58: Good job using lock here so things don't get jumbled up.

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,197 @@
...
+            return new PingResult(combinedExitCode, stringBuilder.ToString());
+        }
+        catch (OperationCanceledException ex)
+        {
+            throw new AggregateException(new TaskCanceledException("Task Ended", ex));
+        }
+    }
+
+
```
AustinH5544, line 81: Looks like you need to implement RunLongRunningAsync()
Here is a hint that Kevin gave. Use the Task.Factory.StartNew.

return Task.Factory.StartNew(() =>
{
    // Implementation in here
}, token, TaskCreationOptions.LongRunning, TaskScheduler.Current);

Assignment/PingProcess.cs
```diff
@@ -1,173 +1,197 @@
...
+                lock (objectlock)
+                {
+                    if (!string.IsNullOrEmpty(result.StdOutput))
+                    {
+                        stringBuilder.AppendLine(result.StdOutput.Trim());
+                    }
+                }
+                return result.ExitCode;
+
```
AustinH5544, line 67: May want to consider returning a PingResult object because it would contain the exit code and output.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+public class PingProcessTests
+{
+    PingProcess Sut { get; set; } = new();
+
+    [TestInitialize]
+    public void TestInitialize()
+    {
+        Sut = new();
+    }
+
```
AustinH5544, line 21: We had to add -4 4 to the destination of the ping to get these to work on the PR. 

[TestMethod]
public void Start_PingProcess_Success()
{
    Process process = Process.Start("ping", "localhost -c 4");
    process.WaitForExit();
    Assert.AreEqual<int>(0, process.ExitCode);
}

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+    [TestMethod]
+    public void Start_PingProcess_Success()
+    {
+        Process process = Process.Start("ping", "localhost");
+        process.WaitForExit();
+        Assert.AreEqual<int>(0, process.ExitCode);
+    }
+
```
AustinH5544, line 29: We couldn't get these next 2 to work even with the -c 4

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+        Assert.AreEqual<int>(1, exitCode);
+    }
+
+    [TestMethod]
+    public void Run_CaptureStdOutput_Success()
+    {
+        PingResult result = Sut.Run("localhost");
+        AssertValidPingOutput(result);
+    }
+
```
AustinH5544, line 57: The instructions say for public void RunTaskAsync_Success(string address) to use  "localhost" but do "localhost -c 4".
You may be able to leave these other ones but add -c 4. I don't know if it will work correctly though.

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+        // Test Sut.RunTaskAsync("localhost");
+
+        //Act
+        Task<PingResult> result = Sut.RunTaskAsync(address);
+        result.Wait();
+
+        //Assert
+        Assert.AreEqual(0, result.Result.ExitCode);
+        Assert.IsNotNull(result.Result);
+
```
AustinH5544, line 74: We could not get the WildCardPattern to work but you could also make sure the StdOutput is not null or white space:

Assert.IsFalse(string.IsNullOrWhiteSpace(result.StdOutput));

Assignment.Tests/PingProcessTests.cs
```diff
@@ -1,152 +1,201 @@
...
+        // Test Sut.RunAsync("localhost");
+        AssertValidPingOutput(result);
+    }
+
+
+
+    [TestMethod]
+    [ExpectedException(typeof(AggregateException))]
+    public void RunAsync_UsingTplWithCancellation_CatchAggregateExceptionWrapping()
+    {
```
AustinH5544, line 110: Looks like it wants you to simplify this expression:

CancellationTokenSource cts = new(true);

AustinH5544 (commented):
Instructions
First implement public void RunTaskAsync_Success() test method to test PingProcess.RunTaskAsync() using "localhost". ❌ // Use localhost
First implement the public void RunAsync_UsingTaskReturn_Success() test method to test PingProcess.RunAsync() using "localhost" without using async/await. ❌ // Use localhost
Complete/fix AND test async public Task<PingResult> RunAsync(IEnumerable<string> hostNameOrAddresses, CancellationToken cancellationToken = default) which executes ping for an array of hostNameOrAddresses (which can all be "localhost") in parallel, adding synchronization if needed. ✔ NOTE:
StdOutput must have all the ping output returned (no lines can be missing) even though intermingled. ✔ // I would recommend returning a PingResult though 
Implement AND test public Task<int> RunLongRunningAsync(ProcessStartInfo startInfo, Action<string?>? progressOutput, Action<string?>? progressError, CancellationToken token) using Task.Factory.StartNew() and invoking RunProcessInternal with a TaskCreation value of TaskCreationOptions.LongRunning and a TaskScheduler value of TaskScheduler.Current. Returning a Task<PingResult> is also okay. NOTE: This method does NOT use Task.Run. ❌ // Not impemented from what I could see
Extra Credit
Test and implement PingProcess.RunAsync(System.IProgress<T> progess) so that you can capture the output as it occurs rather than capturing the output only when the process finishes. ❌
Fundamentals
Place all shared project items into a Directory.Build.targets file.✔

Assignment/PingProcess.cs
1Kmiller1: Thanks for the tip!

KennethWhite (commented):
Overall good work! A few issues to work on, particularly around how we want to handle exceptions in C#. A few minor testing issues that could be cleaned up/improved, but nice job with all the Task work!

